PR #17, opened 2024-08-08
feat: cleanup generated folder

jpudysz:
When I was experimenting with some nitrogen options, I noticed that the generated folder retains old directories and doesn't remove them when there is no spec.

This small PR ensures that the generated folder will be removed recursively whenever the user runs script. Additionally, if the user decides to remove specs and runs the script again, the folder will be empty.

Before:

Example if I try to generate code from swift / kotlin to C++
image

After:

image

Also there is a small indicator that folder has been removed:
image


mrousavy:
Obsolete now

PR #34, opened 2024-08-14
feat: Create `+ResultTypes` for Swift again as we now use C++ types

mrousavy:
Wraps all results of Swift functions in a `Name_method_Result` type that looks like this:

```swift
@frozen
public enum HybridSwiftKotlinTestObjectSpecCxx_callCallback_Result {
  case successful(Double)
  case failure(error: String)
}
```

This way we catch errors and propagate them to C++ safely, as a `throw` in Swift cannot be caught in C++ but instead would crash the entire app.

The problem is now that this won't build tho...


mrousavy:
already done

PR #116, opened 2024-09-11
fix: Handle all discriminated unions properly

boorad:
Nitrogen doesn't handle all [discriminated unions](https://www.typescriptlang.org/docs/handbook/typescript-in-5-minutes-func.html#discriminated-unions) properly.

When adding a few examples to TestObject.nitro.ts spec, we see the following error:

```
❌  Failed to generate spec for TestObject! Error: String literal "one" cannot be represented in C++ because it is ambiguous between a string and a discriminating union enum.
```


packages/nitrogen/src/syntax/createType.ts
```diff
@@ -257,10 +257,6 @@ export function createType(type: TSMorphType, isOptional: boolean): Type {
       throw new Error(
         `Anonymous objects cannot be represented in C++! Extract "${type.getText()}" to a separate interface/type declaration.`
       )
-    } else if (type.isStringLiteral()) {
```
boorad, line 260: @mrousavy I feel like this was here for a reason.  While these changes "work" and create generated files without error, I sense they are breaking another part of nitro handling strings/enums/etc.

mrousavy:
android build failing - i'll check this out when i'm sober

mrousavy:
Fixed this in https://github.com/mrousavy/nitro/pull/130! :)

PR #122, opened 2024-09-14
fix: template cpp-adapter.cpp for Android

TheRogue76:
Think this might have been missed in the template. Still trying to get this up and running locally

Edit: Looking at the CMake file, i am seeing the Hybrid test object, but no `JHybridKotlinTestObjectSpec`. Guessing that will be part of the autogenerated spec. Removed the HybridImageFactory though as that looks like it comes straight from the sample proj

### DIFF
--- a/packages/template/android/src/main/cpp/cpp-adapter.cpp
+++ b/packages/template/android/src/main/cpp/cpp-adapter.cpp
@@ -2,8 +2,6 @@
 
 #include "JFunc_void_Person.hpp"
 #include "JFunc_void_std__string.hpp"
-#include "JHybridImageFactorySpec.hpp"
-#include "JHybridImageSpec.hpp"
 #include "JHybridKotlinTestObjectSpec.hpp"
 
 #include "HybridTestObject.hpp"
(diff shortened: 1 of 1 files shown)


packages/template/android/src/main/cpp/cpp-adapter.cpp
```diff
@@ -2,15 +2,13 @@
...
-#include "JHybridImageFactorySpec.hpp"
-#include "JHybridImageSpec.hpp"
 #include "JHybridKotlinTestObjectSpec.hpp"
 
 #include "HybridTestObject.hpp"
 #include <NitroModules/HybridObjectRegistry.hpp>
 #include <fbjni/fbjni.h>
 
-using namespace margelo::nitro::image;
+using namespace margelo::nitro::<<androidCxxLibName>>;
```
TheRogue76, line 11: One solution could also be to not go this deep in the namespace in the template, like we have on iOS. Then there is no need for this

mrousavy: This then could cause nameclashes :)

PR #129, opened 2024-09-14
feat: Implement AnyMap in Java

mrousavy:
This might be faster... Not sure yet


mrousavy:
lots of code for nothing - this is not faster. In fact, it is 50% slower than before when creating maps, and **5x slower** when passing maps around (JS -> native, and native -> JS).

Issue #125, opened 2024-09-14
Combining optionals with union doesn't always work


grabbou:
For the following signatures:
-  `(event?: Type | TypeB)` 
- `(event: Type | TypeB)`,

Nitrogen will generate enum that are required

I tried to workaround with empty interface, but that doesn't work (see #123)

mrousavy:
Does `event: Type | TypeB | undefined` work?

mrousavy:
Fixed this in https://github.com/mrousavy/nitro/pull/130! :)

PR #130, opened 2024-09-16
fix: Fix optional enums in Nitrogen

mrousavy:
Fixes enum parsing by properly detecting the non-nullable type in unions.

This change now allows these APIs to work:

```ts
(event?: Type | TypeB)
(event: Type | TypeB)
```

Previously that broke.

- Fixes https://github.com/mrousavy/nitro/issues/125
- Fixes https://github.com/mrousavy/nitro/pull/116

### DIFF
--- a/packages/nitrogen/src/syntax/createType.ts
+++ b/packages/nitrogen/src/syntax/createType.ts
@@ -209,10 +209,14 @@ export function createType(type: TSMorphType, isOptional: boolean): Type {
       return new EnumType(typename, enumDeclaration)
     } else if (type.isUnion()) {
       // It is some kind of union - either full of strings (then it's an enum, or different types, then it's a Variant)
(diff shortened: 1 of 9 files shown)


packages/nitrogen/src/syntax/createType.ts
```diff
@@ -258,9 +265,7 @@ export function createType(type: TSMorphType, isOptional: boolean): Type {
         `Anonymous objects cannot be represented in C++! Extract "${type.getText()}" to a separate interface/type declaration.`
       )
     } else if (type.isStringLiteral()) {
-      throw new Error(
-        `String literal ${type.getText()} cannot be represented in C++ because it is ambiguous between a string and a discriminating union enum.`
-      )
+      return new StringType()
```
boorad, line 268: Can you move this up above with the `isString()` clause?

PR #136, opened 2024-09-16
fix: pass in generic types and export Specs

itsramiel:
## Summary
The nitrogen page is a step by step guide on running Nitro Codegen but following as it is now leads user to errors.

When defining the Spec, HybridObject requires a generic parameter or else the user would get:
```
        ❌  Failed to generate spec for RNFilePicker! Error: RNFilePicker does not properly extend HybridObject<T> - HybridObject does not have a single generic type argument for platform spec languages.
    ❌  No specs found in RNFilePicker.nitro.ts!
```

So this pr adds the generic type for guide completeness and exports the Spec since it is used in step 6 which I also made obvious in that step that this is the spec defined in .nitro.ts file.

Also the classes are missing the memorySize and hybridContext(for ios) for completeness. I copy pasted them, only to get a cryptic error in the console till i opened xcode. C++ needs to be updated too for memorySize? Not sure

### DIFF
--- a/docs/docs/nitrogen.md
+++ b/docs/docs/nitrogen.md
@@ -105,7 +105,7 @@ Nitrogen parses all TypeScript files that end in `.nitro.ts`.
 For example, let's create `Math.nitro.ts`:
 
 ```ts title="Math.nitro.ts"
-interface Math extends HybridObject {
+export interface Math extends HybridObject<{ ios: 'swift', android: 'kotlin'}> {
   add(a: number, b: number): number
 }
 ```
--- a/docs/docs/nitrogen.md
+++ b/docs/docs/nitrogen.md
@@ -335,7 +335,7 @@ class MainApplication {
 
 ### 6. Initialize the Hybrid Objects
 
-And finally, to initialize `HybridMath` from JS you just need to call `createHybridObject`, passing in the `Math` Spec defined in `Math.nitro.ts` as a generic type parameter:
+And finally, to initialize `HybridMath` from JS you just need to call `createHybridObject`:
 
 ```ts
 export const MathModule = NitroModules.createHybridObject<Math>("Math")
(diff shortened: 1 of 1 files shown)


docs/docs/nitrogen.md
```diff
@@ -105,7 +105,7 @@ Nitrogen parses all TypeScript files that end in `.nitro.ts`.
 For example, let's create `Math.nitro.ts`:
 
 ```ts title="Math.nitro.ts"
-interface Math extends HybridObject {
+export interface Math extends HybridObject<{ ios: 'swift', android: 'kotlin'}> {
```
mrousavy, line 108:
```suggestion
interface Math extends HybridObject<{ ios: 'swift', android: 'kotlin' }> {
```

docs/docs/nitrogen.md
```diff
@@ -203,6 +203,12 @@ To implement `Math` now, you just need to implement the spec:
   <TabItem value="swift" label="Swift" default>
     ```swift title="HybridMath.swift"
     class HybridMath : HybridMathSpec {
+      var hybridContext = margelo.nitro.HybridContext()
+
```
mrousavy, line 207:
```suggestion
```

docs/docs/nitrogen.md
```diff
@@ -327,7 +336,7 @@ class MainApplication {
 
 ### 6. Initialize the Hybrid Objects
 
-And finally, to initialize `HybridMath` from JS you just need to call `createHybridObject`:
+And finally, to initialize `HybridMath` from JS you just need to call `createHybridObject`, passing in the `Math` Spec defined in `Math.nitro.ts` as a generic type parameter:
```
mrousavy, line 339:
```suggestion
And finally, to initialize `HybridMath` from JS you just need to call `createHybridObject`:
```

mrousavy (commented):
looks good, but I have a few suggestions to keep it a bit cleaner - less text is more readable.

PR #137, opened 2024-09-16
fix: fix android placeholders

itsramiel:
## Summary

The `<<androidCxxLibName>>Package.java` has the `image` component package at the very top but instead should be the placeholder.

Also we can add loading the c++ library with the placeholder


mrousavy:
perfect, thank you!

PR #140, opened 2024-09-18
fix: template new library android sync project error

kingsword09:
When I create a new library following the tutorial above in [Creating a Nitro Module](https://mrousavy.github.io/nitro/docs/nitro-modules#creating-a-nitro-module), I get some error in the `Android Studio` `sync project` with errors, so fix these exceptions.

1. Add the location of the `react-native-nitro-module module` to `settings.gradle`:
image

2. Add `classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:$<<androidCxxLibName>>_kotlinVersion"` fix the following problems:
image

3. Upgraded the version of `com.android.tools.build:gradle` because `compileSdkVersion` `34` requires at least version `8.1.1`.

4. use `reactNativeArchitectures=true` fix the following problems:
image

### DIFF
--- a/packages/template/android/build.gradle
+++ b/packages/template/android/build.gradle
@@ -5,7 +5,8 @@ buildscript {
   }
 
   dependencies {
-    classpath "com.android.tools.build:gradle:7.2.1"
+    classpath "com.android.tools.build:gradle:8.1.1"
+    classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:$<<androidCxxLibName>>_kotlinVersion"
   }
 }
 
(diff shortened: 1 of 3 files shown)


mrousavy:
How did you fix your error? why close this PR?

kingsword09:
> How did you fix your error? why close this PR?

There is no fix, I copied the template directly, according to the tutorial to change to my project name after wanting to use Android Studio directly to open the android directory sync project, and found that there will be an error, my above modification is also still a problem, I want to think in the research to see!

kingsword09:
Since it's been a while since I've tried a react native project, I may need to take another look at the documentation for developing a new library directly. I thought it was possible to run it directly in `Android Studio` if I created it directly using the template.

mrousavy:
No this will be the package only - you need to create an example app as well - that can be bare RN or expo.

kingsword09:
> No this will be the package only - you need to create an example app as well - that can be bare RN or expo.

thanks

PR #144, opened 2024-09-23
fix: Make `NativeNitroModules` a code-gen'd TurboModule again

mrousavy:
This has two benefits:

1. The app only builds if the `NativeNitroModules` spec is aligned with the TypeScript types
2. The Java module now receives `ReactApplicationContext`

..but one potential minor downsides:

1. There might be changes in CxxTurboModules in the future that could cause build issues with Nitro. Hopefully no breaking changes


mrousavy:
ah, I forgot that we cannot do the same on Android. On Android, they expect the React Application's CMakeList.txt to compile my library's C++ code for whatever reason.

This complicates things as we cannot autolink or use prefabs anymore, so I can't go that route.

PR #177, opened 2024-10-05
perf: Mark SwiftCxx wrappers `struct` instead of `class` for better performance

mrousavy:
`struct` has better performance than `class`, let's see if this actually works tho


mrousavy:
nvm, we need inheritance here so we cannot use `struct`.

PR #212, opened 2024-10-10
feat: Allow empty Struct in C++

### LINKED ISSUE #123: Nitrogen allows empty interface, which doesn't compile
```
export interface Open {}

export interface Closed {
  code: string
  reason: string
}

export interface Foo extends HybridObject<{ ios: 'swift' }> {
  create(
    callback: (
      type: string,
      event: Open | Closed
    ) => void
  ): WebSocket
}
```

Error below

Screenshot 2024-09-14 at 16 42 19

mrousavy:
I don't know why you would need that, but at least this is better than a compiler error.

Fixes #123


mrousavy:
nvm I think this is just unnecessary complexity. let's just dont allow this.

PR #213, opened 2024-10-10
fix: Manually convert C++ vector to Swift Array

mrousavy:
Replaces 

```swift
array.map({ __item in String(__item) })
```

with

```swift
({
  var __array: [String] = []
  __array.reserveCapacity(array.count)
  for __i in 0...array.count {
    __array.append(String(array[__i]))
  }
  return __array
}())
```

to manually convert vectors from C++ to Swift.

`map` worked fine in Debug, but apparently broke in release. No idea why.

Created an issue for this bug in apple/swift compiler: https://github.com/swiftlang/swift/issues/76949

### DIFF
--- a/.github/workflows/build-ios.yml
+++ b/.github/workflows/build-ios.yml
@@ -77,7 +77,7 @@ jobs:
           -workspace NitroExample.xcworkspace \
           -scheme NitroExample \
           -sdk iphonesimulator \
-          -configuration Debug \
+          -configuration Release \
           -destination 'platform=iOS Simulator,name=iPhone 15' \
           build \
           CODE_SIGNING_ALLOWED=NO | xcpretty"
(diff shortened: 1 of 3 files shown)


mrousavy:
cc @perunt

mrousavy:
This did actually not work. Simply accessing any property of a `std::vector<std::string>` (even `.count`) crashes the Swift compiler **in release**.

Workarounds:
- For Taras' use-case, we just used enums (`ContactKey[]`) instead of `string[]`
- For the case where you actually need `string[]`, you can create a `interface StringHolder { value: string }`, then just use `StringHolder[]`.

I agree that this is not ideal and will do my best to fix this - I can easily build a workaround in C++ and bridge to that for strings only, but ideally we want to keep it simple and have that compiler bug fixed. :)

PR #223, opened 2024-10-14
Fix Hermes location detection for Android example app

perunt:
This PR addresses the build failure occurring during the creation of the release bundle for Android. The error was caused by the inability to determine the Hermes location.


mrousavy:
LGTM thx

PR #226, opened 2024-10-15
feat: Separate SwiftCxx bridge `.hpp` from `.cpp`

mrousavy:
Separates implementation and header declaration in the Swift Cxx bridge.

This should fix recursive include issues.


mrousavy:
superseded by https://github.com/mrousavy/nitro/pull/232

PR #228, opened 2024-10-15
fix: Allow functions as properties again

mrousavy:
This is different from methods. It's an actual property.


mrousavy:
@grabbou

PR #239, opened 2024-10-17
perf: Speed up `std::vector` by using `emplace_back`

mrousavy:


mrousavy:
This is the same as we have a temp lvalue.

PR #241, opened 2024-10-17
feat: Create `new` prototype

mrousavy:
Adds a new API (`getHybridObjectConstructor(...)`) that gives the user constructor functions for their Hybrid Objects.

This then allows the library users to use the JS `new` operator, and also supports `instanceof` now!

Before:

```ts
const image1 = NitroModules.createHybridObject<Image>('Image')
const image2 = NitroModules.createHybridObject<Image>('Image')

const imagePrototype = Object.getPrototypeOf(image1)
const isImage = Object.getPrototypeOf(image1) === imagePrototype
```

After:

```ts
const HybridImage = getHybridObjectConstructor<Image>('Image')

const image1 = new HybridImage()
const image2 = new HybridImage()
const isImage = image1 instanceof HybridImage
```

One limitation: If you call `instanceof HybridImage` before even calling `new HybridImage()` once in the entire app lifetime, it will lazily call `new HybridImage()` under the hood to create a temporary `HybridImage()` instance once just so that the prototype is known. This happens only once, and I think it's a fair restriction - I can't think of any case where `instanceof` is called beforehand, and even if - it's just _one_ extra `new` call internally.


mrousavy:
btw., I think this has to stay in JS. In Hermes, we cannot call `new [nativeFunc]()` if `[nativeFunc]` is a `jsi::HostFunction`. bug report: https://github.com/facebook/hermes/issues/524

PR #242, opened 2024-10-18
fix(android): fix code gen Onload.{hpp,cpp,kt}

ngocle2497:
Fix auto gen *OnLoad.{cpp,hpp,kt}


mrousavy:
ah, nice catch! thank you :)

PR #249, opened 2024-10-21
feat: change iosModulename to iosModuleName

ngocle2497:
Camel case iosModuleName


mrousavy:
Good stuff, thanks!

mrousavy:
Ah I just realized that this is a breakign change

PR #251, opened 2024-10-21
feat: Allow sync callbacks in `JSIConverter<T>`

mrousavy:
Allows synchronous functions in `JSIConverters`, which allows synchronously and immediately calling a JS callback without "scheduling" it.

This obviously only works from the JS Thread.


mrousavy:
Closing in favor of https://github.com/mrousavy/nitro/pull/564 - which now finally works. Two separate types, fast, perfect-forwarding, and Swift/Kotlin compatible.

There are two (unavoidable) downsides with the approach though:
1. Whether a callback will be called synchronously or asynchronously has to be decided upfront in the TypeScript spec. You cannot choose to either call it sync or async inside your native method later on - you have to know upfront when defining the specs.
2. A callback that returns `void` is ambiguous between sync and async - so it is async (shoot and forget) by default. If you need a sync callback that doesn't return anything, just return a dummy value like a `boolean`.

PR #253, opened 2024-10-22
test: Add optional enum tests

mrousavy:
Tests this:

```ts
export type SelectMode = 'single' | 'multiple'

export interface Options {
    selectMode?: SelectMode
}
```

### DIFF
--- a/example/src/getTests.ts
+++ b/example/src/getTests.ts
@@ -248,6 +248,23 @@ export function getTests(
           testObject.optionalHybrid = undefined
         })
     ),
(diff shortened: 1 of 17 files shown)


patrickkabwe:
@mrousavy thanks for this fix. I wanted to check in if this PR will also fix the case below: 

```js
export type SelectMode = 'single' | 'multiple'
export type OtherType = 'somevalue' | 'someothervalue'

export interface Options {
    selectMode?: SelectMode
    otherMode?: OtherType
}
```

mrousavy:
i don't get it, what is different in your code than in mine?

PR #267, opened 2024-10-30
chore: Re-structure project to use global ESLint, Prettier and Typescript configs

chrispader:
Re-structures the project by moving configuration files for ESLint, Typescript and Prettier to a global `config` folder.

All nitro implementation sub-packages (`packages/*`, `example`; *except* `docs` and `template`) will extend the global configs and apply incremental changes locally.

All packages will use the same version of these toolsets and there are global Bun scripts for linting and typechecking... 

This PR also cleans up the `package.json` of the root and sub-packages by extracting the ESLint config and implementing a universal order of properties.

Additionally, the PR improves the release workflow by enforcing consistent version numbers across all workspace packages. The PR also lays the fundament for a global release workflow, where all packages are released in one take.


chrispader:
Ready for review :) cc @mrousavy

package.json
```diff
@@ -38,6 +56,18 @@
...
+          {
+            "file": "packages/react-native-nitro-image/package.json",
+            "path": "version"
+          },
+          {
+            "file": "example/package.json",
+            "path": "version"
+          }
+        ]
+      },
```
chrispader, line 70: This just also bumps the versions in `react-native-nitro-image` and `example`

mrousavy:
Run nitrogen

package.json
mrousavy: nice!

packages/react-native-nitro-modules/package.json
```diff
@@ -2,9 +2,9 @@
   "name": "react-native-nitro-modules",
   "version": "0.14.0",
   "description": "Insanely fast native C++, Swift or Kotlin modules with a statically compiled binding layer to JSI.",
-  "main": "lib/index",
-  "module": "lib/index",
-  "types": "lib/index.d.ts",
```
mrousavy, line 7: oh interesting - was this technically broken before?

packages/react-native-nitro-modules/src/turbomodule/NativeNitroModules.ts
```diff
@@ -42,7 +42,7 @@ declare global {
 }
 
 export function isRuntimeAlive() {
-  const cache = global.__nitroModulesJSICache
-  const dispatcher = global.__nitroDispatcher
```
mrousavy, line 46: Why? I don't think this makes sense.

packages/react-native-nitro-modules/package.json
mrousavy, line 7: oh interesting - was this technically broken before?

scripts/clang-format.sh
```diff
@@ -14,7 +14,7 @@ CPP_DIRS=(
 if which clang-format >/dev/null; then
   DIRS=$(printf "%s " "${CPP_DIRS[@]}")
   find $DIRS -type f \( -name "*.h" -o -name "*.hpp" -o -name "*.cpp" -o -name "*.m" -o -name "*.mm" -o -name "*.c" \) -print0 | while read -d $'\0' file; do
-    clang-format -style=file:./.clang-format -i "$file"
+    clang-format -style=file:.clang-format -i "$file"
```
mrousavy, line 17: huh? This must be in `config`

mrousavy (changes requested):

scripts/clang-format.sh
chrispader: The problem is, that you cannot configure the GH action to use a different `.clang-format` file path. In the script it would work, but not in the GH workflow

chrispader: https://github.com/jidicula/clang-format-action

packages/react-native-nitro-modules/package.json
chrispader: yes, the files couldn't be resolved before

packages/react-native-nitro-modules/src/turbomodule/NativeNitroModules.ts
chrispader: Using the `@tsconfig/react-native` config enforces to use `globalThis`. `globalThis` can be used as a drop-in replacement for `global``
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/globalThis

scripts/clang-format.sh
mrousavy: Ah true. There was a PR to add this tho...

chrispader: fixed the path typo though

chrispader: > Ah true. There was a PR to add this tho...

i'll check it out in 30mins

packages/react-native-nitro-modules/src/turbomodule/NativeNitroModules.ts
chrispader: Screenshot 2024-11-04 at 13 09 02

mrousavy: sounds good, nice work 👍

mrousavy (approved):
looks perfect to me now. just a quick q; how does the release process work now? 
Just `bun release` in root?

chrispader:
> looks perfect to me now. just a quick q; how does the release process work now? Just `bun release` in root?

Technically yes, but i couldn't test this without actually triggering the publish. The `workspaces` prop in `<root>/package.json` is sorted by dependency on other workspace packages, e.g. `react-native-nitro-modules` is built and released before `nitrogen`, which is built before `react-native-nitro-image`, and so on.

`version` props and cross-referencing dependency versions will also be updated in the release flow.

mrousavy:
Only react-native-nitro-modules and nitrogen are actually published to npm.

chrispader:
> Only react-native-nitro-modules and nitrogen are actually published to npm.

Yes, still the case. Just wanted to demonstrate the workspace package order

PR #286, opened 2024-11-05
fix: Fix JNI destructor being potentially called from a wrong Thread

mrousavy:
Fixes an issue where the destructor of a C++ JNI object could potentially be called from a Thread that isn't attached to the JNI environment (like a JS garbage collector thread)

- Fixes https://github.com/callstackincubator/react-native-fast-ws/issues/15

cc @grabbou


grabbou:
Yeah, that seems to do the trick! Please release thank you

PR #290, opened 2024-11-05
docs: Update to docusaurus 3.6.0

mrousavy:


mrousavy:
hmm, it doesn't build because of a `chalk` import?

image

PR #294, opened 2024-11-05
chore: Hoist RN, React, Jest and @types in root

chrispader:
Moves explicit package versions to `<root>/package.json` and only uses `*` wildcards in workspace packages.

This makes it easier to bump package versions in the future


mrousavy:
on second thought; could we also use a fixed version (`^0.75.2`) in `example/`, but leave it at wildcard (`*`) for stuff like `packages/react-native-nitro-modules` where it theoretically doesn't matter which RN version it uses as it is a peer/dev-dependency only?

PR #298, opened 2024-11-06
fix: Import Nitro core in Kotlin struct and function

mrousavy:
When using stuff from Nitro Core like `ArrayBuffer` in a Kotlin struct or function, the import was missing. This PR fixes that.

Also fixes nitro-codegen not locally rebuilding when running the `specs` command


mrousavy:
cc @grabbou

PR #299, opened 2024-11-06
fix: Throw if a reserved keyword is used in nitrogen

mrousavy:
That's `core`, `NitroModules`, and `nitro`


grabbou:
Also, happens when using `protocol` :D

mrousavy:
This is only for namespace rn, not really for struct, enum, hybridobject, function, parameter, or property names yet

PR #302, opened 2024-11-06
chore: Automatic release flow through release script

chrispader:
This PR adds a release script that will automatically publish `react-native-nitro-modules`, then `nitrogen` and then create a Git commit and GH release.

To trigger the release process, just run `bun release <version>` or `bun release --increment <major|minor|patch>` in the top-level root folder.

If preferred, the sub-packages can still also be released manually by running `bun release` in one of the packages


example/package.json
```diff
@@ -11,6 +11,7 @@
     "typecheck": "tsc --noEmit",
     "lint": "eslint \"**/*.{js,ts,tsx}\" --fix",
     "lint-ci": "eslint \"**/*.{js,ts,tsx}\" -f @jamesacarr/github-actions",
+    "bundle-install": "cd ios && bundle install",
```
mrousavy, line 14: Shouldn't this be executed in `example/` instead of `example/ios/`?

chrispader: True, both works, but this is better

mrousavy (approved):

PR #315, opened 2024-11-10
Fix built package autolinking

andreazllin:
Currently the package.json doesn't include the nitrogen folder which is needed to execute the autolinking


mrousavy:
Nice catch!

PR #320, opened 2024-11-11
Update using-nitro-in-your-app.md

vilindberg:
I would add this since I actually wasted like 5 minutes on it. Just copied and pasted like a true lazy dev.


mrousavy:
I would argue that imports are not required to be put into the docs.

Isn't it obvious that this has to be imported? There's a red squiggly line in every editor if you don't import it

vilindberg:
Yeah, it was kind of silly of me to miss it, but I did. IMO it's nice when the copy code-click + paste flow is seamless.

PR #321, opened 2024-11-12
fix: Fix hard crash on Android in Prototype tests in release

mrousavy:
I honestly don't know why this crashes.

But now we check for `hasNativeState()` also in release builds, not only debug. This will then `throw` if the function is called on an _unbound_ object (no `this`/`NativeState`).

I can only explain this being called by `console.log`'ing an _unbound_ object (so just the Prototype) - which has an overridden `toString` function but that requires it to be _bound_.
I guess if `console.log` calls `toString()` and `toString()` throws, it will catch the error internally.. idk man.

no real production app will call `toString()` on a prototype **in release**, so I'm not sure if this is the right approach.

### DIFF
--- a/packages/react-native-nitro-modules/android/src/main/cpp/platform/NitroLogger.cpp
+++ b/packages/react-native-nitro-modules/android/src/main/cpp/platform/NitroLogger.cpp
@@ -27,9 +27,11 @@ int levelToAndroidLevel(LogLevel level) {
 }
 
 void Logger::nativeLog(LogLevel level, const char* tag, const std::string& message) {
+#ifdef NITRO_DEBUG
   int logLevel = levelToAndroidLevel(level);
   std::string combinedTag = "Nitro." + std::string(tag);
   __android_log_print(logLevel, combinedTag.c_str(), "%s", message.c_str());
+#endif
 }
 
 } // namespace margelo::nitro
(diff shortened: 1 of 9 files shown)


mrousavy:
Wait no, huh??

This code here works:

Screenshot 2024-11-12 at 11 43 44

and this code here crashes:

Screenshot 2024-11-12 at 11 44 11

..same as if I would just not call `hasNativeState` at all, this also crashes:

Screenshot 2024-11-12 at 11 44 22



So what makes it work? The `assert(..)` can't be hit, can it? It would crash the app...

mrousavy:
Okay some more insights:

1. This code here crashes:
    ```cpp
    auto nativeState = object.getNativeState(runtime);
    ```
2. This code here also crashes:
    ```cpp
    object.hasNativeState(runtime);
    auto nativeState = object.getNativeState(runtime);
    ```
3. This code here **WORKS**, but it doesn't assert/crash the app???:
    ```cpp
    if (!object.hasNativeState(runtime)) {
      assert(false);
    }
    auto nativeState = object.getNativeState(runtime);
    ```
4. This code here **WORKS**:
    ```cpp
    if (!object.hasNativeState(runtime)) {
       value.toString(runtime);
    }
    auto nativeState = object.getNativeState(runtime);
    ```

So maybe the `object` is somehow flattened or lazy evaluated in a sense that `NativeState` isn't properly available?

I am still kinda baffled that the `assert` does not crash the app, maybe `object.hasNativeState(...)` is not the same as the internal assert of `object.getNativeState(...)`?

cc @hannojg @tmikov

mrousavy:
Created an issue in the Hermes repo for now: https://github.com/facebook/hermes/issues/1564

mrousavy:
I think this can be fixed by just adjusting the code in the example app. No need to slow down all function calls for users by leaving this check in release builds

Issue #218, opened 2024-10-13
Build fails with these errors 'folly/folly-config.h' file not found and Could not build Objective-C module 'React' 


patrickkabwe:
### How were you trying to build the app?

After running `bun ios` to build the app I am getting the errors(PLEASE see the full build logs section).

Tried running the app from Xcode (16) I am getting the following(SEE ATTACHED SCREENSHOT):

<img width="1504" alt="image" src="https://github.com/user-attachments/assets/07881248-d22d-4364-995a-d00e1054961e">

### Full build logs

```tsx
$ react-native run-ios --simulator='iPhone 16'
info A dev server is already running for this project on port 8081.
info Found Xcode workspace "NitroInAppBrowser.xcworkspace"
info Building (using "xcodebuild -workspace NitroInAppBrowser.xcworkspace -configuration Debug -scheme NitroInAppBrowser -destination id=5BE3F125-2E7F-446C-99AE-85585656EA9D")
▸ Processing ResourceBundle-React-Core_privacy-React-Core-Info.plist
▸ Processing ResourceBundle-React-cxxreact_privacy-React-cxxreact-Info.plist
▸ Running script '[CP-User] [Hermes] Replace Hermes for the right configuration, if needed'
▸ Processing ResourceBundle-glog_privacy-glog-Info.plist
▸ Processing ResourceBundle-boost_privacy-boost-Info.plist
▸ Processing ResourceBundle-RCT-Folly_privacy-RCT-Folly-Info.plist
▸ Running script '[CP] Copy XCFrameworks'
▸ Compiling NSDataBigString.mm
▸ Compiling RCTCallInvoker.mm
▸ Compiling RCTFont.mm
▸ Compiling RCTBundleURLProvider.mm
▸ Compiling RCTDefaultCxxLogFunction.mm
▸ Compiling RCTFollyConvert.mm
▸ Compiling RCTUIManagerObserverCoordinator.mm
▸ Compiling RCTSurfaceView.mm
▸ Compiling RCTSurfaceSizeMeasureMode.mm
▸ Compiling RCTSurfaceRootView.mm
▸ Compiling RCTSurfaceHostingView.mm
▸ Compiling RCTSurfaceHostingProxyRootView.mm
▸ Compiling RCTSurface.mm
▸ Compiling RCTPerformanceLogger.mm
▸ Compiling RCTPausedInDebuggerOverlayController.mm
▸ Compiling RCTPackagerConnection.mm
[...]

PR #333, opened 2024-11-14
support USE_FRAMEWORKS=static

baveku:
Fix issue: #219 #218


packages/nitrogen/src/autolinking/ios/createPodspecRubyExtension.ts
```diff
@@ -61,6 +61,9 @@ def add_nitrogen_files(spec)
     "SWIFT_OBJC_INTEROP_MODE" => "objcxx",
     # Enables stricter modular headers
     "DEFINES_MODULE" => "YES",
+		# Update Search Paths
+		"USER_HEADER_SEARCH_PATHS" => '"\${PODS_CONFIGURATION_BUILD_DIR}/\${SWIFT_MODULE_NAME}/\${SWIFT_MODULE_NAME}.framework/Headers"',
+		"GCC_PREPROCESSOR_DEFINITIONS" => "FOLLY_NO_CONFIG=1 FOLLY_CFG_NO_COROUTINES=1"
```
mrousavy, line 66:
```suggestion
    # Update Search Paths
    "USER_HEADER_SEARCH_PATHS" => '"\${PODS_CONFIGURATION_BUILD_DIR}/\${SWIFT_MODULE_NAME}/\${SWIFT_MODULE_NAME}.framework/Headers"',
    "GCC_PREPROCESSOR_DEFINITIONS" => "FOLLY_NO_CONFIG=1 FOLLY_CFG_NO_COROUTINES=1"
```

packages/react-native-nitro-image/cpp/HybridTestObjectCpp.cpp
```diff
@@ -7,7 +7,7 @@
 
 #include "HybridTestObjectCpp.hpp"
 #include <NitroModules/AnyMap.hpp>
-#include <NitroModules/NitroLogger.hpp>
```
mrousavy, line 10: Why?

packages/react-native-nitro-image/nitrogen/generated/ios/NitroImage+autolinking.rb
```diff
@@ -54,5 +54,8 @@ def add_nitrogen_files(spec)
     "SWIFT_OBJC_INTEROP_MODE" => "objcxx",
     # Enables stricter modular headers
     "DEFINES_MODULE" => "YES",
+		# Update Search Paths
+		"USER_HEADER_SEARCH_PATHS" => '"${PODS_CONFIGURATION_BUILD_DIR}/${SWIFT_MODULE_NAME}/${SWIFT_MODULE_NAME}.framework/Headers"',
+		"GCC_PREPROCESSOR_DEFINITIONS" => "FOLLY_NO_CONFIG=1 FOLLY_CFG_NO_COROUTINES=1"
```
mrousavy, line 59:
```suggestion
    # Update Search Paths
    "USER_HEADER_SEARCH_PATHS" => '"${PODS_CONFIGURATION_BUILD_DIR}/${SWIFT_MODULE_NAME}/${SWIFT_MODULE_NAME}.framework/Headers"',
    "GCC_PREPROCESSOR_DEFINITIONS" => "FOLLY_NO_CONFIG=1 FOLLY_CFG_NO_COROUTINES=1"
```

packages/react-native-nitro-modules/NitroModules.podspec
```diff
@@ -53,6 +53,8 @@ Pod::Spec.new do |s|
     "SWIFT_OBJC_INTEROP_MODE" => "objcxx",
     # Enables stricter modular headers
     "DEFINES_MODULE" => "YES",
+		# Enables static library,
+		"GCC_PREPROCESSOR_DEFINITIONS" => "$(inherited) FOLLY_NO_CONFIG FOLLY_CFG_NO_COROUTINES"
```
mrousavy, line 57:
```suggestion
    # Enables static library,
    "GCC_PREPROCESSOR_DEFINITIONS" => "$(inherited) FOLLY_NO_CONFIG FOLLY_CFG_NO_COROUTINES"
```

mrousavy (changes requested):
Hey - thanks for your PR.

A few things;

1. Watch indentation. This is not correct, I adjusted it.
2. I'm not really sure if this should be in the autogenerated podspec extension. I would argue that this is something the user has to put in their podspec. Especially the FOLLY part which might change over time - Nitro does not have anything to do with Folly, this is something the user should configure in their Podspec imo.

mrousavy:
In my opinion, this shouldn't go in the `+autolinking.rb` file.

This should go into `NitroModules.podspec` and the template's `.podspec`, but not in the autogenerated part. It's something that might change over time, and it's also an external factor (folly) - not really my fault that this breaks. I don't think going out of my way to fix a bug in another library is the right idea.

baveku:
> In my opinion, this shouldn't go in the `+autolinking.rb` file.
> 
> This should go into `NitroModules.podspec` and the template's `.podspec`, but not in the autogenerated part. It's something that might change over time, and it's also an external factor (folly) - not really my fault that this breaks. I don't think going out of my way to fix a bug in another library is the right idea.

That's right, I'll reference other libraries instead of editing autolink.
```
folly_config = get_folly_config()
folly_compiler_flags = folly_config[:compiler_flags]
folly_version = folly_config[:version]
boost_compiler_flags = '-Wno-documentation'
```

baveku:
Hey @mrousavy,
First, I update umbera to support static, we don't need use HEADER_SEARCH_PATHS
```
// Include Swift defined types
#if __has_include("NitroImage-Swift.h")
// This header is generated by Xcode/Swift on every app build.
// If it cannot be found, make sure the Swift module's name (= podspec name) is actually "NitroImage".
#include "NitroImage-Swift.h"
#elif __has_include(<NitroImage/NitroImage-Swift.h>)
#include <NitroImage/NitroImage-Swift.h>
#else
#error NitroImage's autogenerated Swift header cannot be found! Make sure the Swift module's name (= podspec name) is actually "NitroImage", and try building the app first.
#endif
```
Seconds, I don't found new ways to support build static so I will edit podspec to support GCC_PREPROCESSOR_DEFINITIONS
```
$gcc_shared_flags = "$(inherited) FOLLY_NO_CONFIG FOLLY_CFG_NO_COROUTINES"
gcc_flags = ''
$use_frameworks = ENV['USE_FRAMEWORKS']
if $use_frameworks
	gcc_flags = $gcc_shared_flags
end
...

s.pod_target_xcconfig = {
		"GCC_PREPROCESSOR_DEFINITIONS" => gcc_flags}
}
```
Hope we can use this solution for USE_FRAMEWORKS=static

mrousavy:
Thanks for your PR, unfortunately there were a few things missing here and you didn't tell me why you removed the NitroLogger header and just closed the discussion.

Really appreciate you taking the time to fix this, but I created another PR that takes this over the finish line and now builds successfully: https://github.com/mrousavy/nitro/pull/379

Thanks again for your PR and sorry for not merging yours!

baveku:
> Thanks for your PR, unfortunately there were a few things missing here and you didn't tell me why you removed the NitroLogger header and just closed the discussion.
> 
> Really appreciate you taking the time to fix this, but I created another PR that takes this over the finish line and now builds successfully: #379
> 
> Thanks again for your PR and sorry for not merging yours!

You don't need to worry about that. My team will continue to support and make all libraries with Nitro.
Let's go.

PR #338, opened 2024-11-16
fix: Call `ordinal()` method instead of trying to access field which doesn't exist

mrousavy:
Previously, we tried to access `.ordinal` as a field on an enum.

Apparently this doesn't exist, and `ordinal()` is a method on an enum. So this PR now changes this to call that method.

One thing I am wondering is; how did this work before? Afaik we have those test cases... 🤔 

1. We get an enum from a property here: https://github.com/mrousavy/nitro/blob/d76d0c02621d4183544d2a8af5b8b334318b4551/packages/react-native-nitro-image/src/specs/TestObject.nitro.ts#L52
2. We get an array of enums here: https://github.com/mrousavy/nitro/blob/d76d0c02621d4183544d2a8af5b8b334318b4551/packages/react-native-nitro-image/src/specs/TestObject.nitro.ts#L64
3. We get an enum from a method here: https://github.com/mrousavy/nitro/blob/d76d0c02621d4183544d2a8af5b8b334318b4551/packages/react-native-nitro-image/src/specs/TestObject.nitro.ts#L80

cc @Szymon20000 - didn't those tests work? In the example app they're green...?

### DIFF
--- a/packages/nitrogen/src/syntax/kotlin/KotlinEnum.ts
+++ b/packages/nitrogen/src/syntax/kotlin/KotlinEnum.ts
@@ -58,8 +58,8 @@ namespace ${cxxNamespace} {
     [[maybe_unused]]
     ${enumType.enumName} toCpp() const {
       static const auto clazz = javaClassStatic();
-      static const auto fieldOrdinal = clazz->getField<int>("ordinal");
-      int ordinal = this->getFieldValue(fieldOrdinal);
+      static const auto ordinalMethod = clazz->getMethod<int()>("ordinal");
+      int ordinal = ordinalMethod(self());
       return static_cast<${enumType.enumName}>(ordinal);
     }
 
--- a/packages/react-native-nitro-image/nitrogen/generated/android/c++/JImageFormat.hpp
+++ b/packages/react-native-nitro-image/nitrogen/generated/android/c++/JImageFormat.hpp
@@ -28,8 +28,8 @@ namespace margelo::nitro::image {
     [[maybe_unused]]
     ImageFormat toCpp() const {
       static const auto clazz = javaClassStatic();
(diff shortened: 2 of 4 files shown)


mrousavy:
hm - I just tested this again (in https://github.com/mrousavy/nitro/pull/346), and it seems like `getFieldValue("ordinal")` works as expected for me.
Again, this would've failed in the tests already, we use this property.

So I think this is not a bug - maybe there was something else that was broken in your app @Szymon20000.

image

mrousavy:
Ah there's why this works:
image


In Kotlin, enums have a `ordinal` **field**. In Java, enums have a `ordinal()` **method**.

Since we use Kotlin, we're good.

mrousavy:
@shovel-kun which Kotlin version are you using? This is weird, `ordinal` is definitely a field in my test cases. It works in the example app.

mrousavy:
@shovel-kun can you please test this PR? https://github.com/mrousavy/nitro/pull/360

This is a better solution in my opinion since it has no performance impact. The PR here has one, as we switch from field access to method call.

shovel-kun:
@mrousavy 

Tested the PR and it is working for me.

Not sure how to check what Kotlin version I'm using, I think 2.0.0

```gradle
    classpath "com.android.tools.build:gradle:7.2.1"
    classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:2.0.21"
    classpath "org.jetbrains.kotlin:kotlin-serialization:2.0.21"
    
    …
    
    apply plugin: "com.android.library"
    apply plugin: 'org.jetbrains.kotlin.android'
    apply from: '../nitrogen/generated/android/Tachiyomi+autolinking.gradle'

```

mrousavy:
great, thx

PR #340, opened 2024-11-17
repro: Add test where Promise.all checks if they can run in parallel

grabbou:
On Android, when there are two promises, second promise will wait for the first one to resolve. 

In this example, the resolution should be:
- Promise.all(foo(), bar()) called, both functions are running
- foo() is pending
- bar() is pending
- bar gets resolved
- 2 seconds later
- foo() gets resolved
- Promise.all gets resolved

### DIFF
--- a/example/src/getTests.ts
+++ b/example/src/getTests.ts
@@ -660,6 +660,26 @@ export function getTests(
     createTest('promiseThrows() throws', async () =>
       (await it(() => testObject.promiseThrows())).didThrow()
     ),
+    createTest('twoPromises', async () =>
+      (
+        await it(async () => {
+          let fooResolved: boolean | undefined
+          let barResolvedBeforeFoo: boolean | undefined
+          const foo = async () => {
+            await testObject.wait(2)
+            fooResolved = true
+          }
+          const bar = async () => {
+            await testObject.createArrayBufferAsync()
+            barResolvedBeforeFoo = !fooResolved
+          }
+          await Promise.all([foo(), bar()])
+          return barResolvedBeforeFoo
+        })
+      )
+        .didNotThrow()
+        .equals(true)
+    ),
 
     // Callbacks
     createTest('callCallback(...)', async () =>
--- a/example/src/getTests.ts
+++ b/example/src/getTests.ts
@@ -660,21 +660,15 @@ export function getTests(
     createTest('promiseThrows() throws', async () =>
       (await it(() => testObject.promiseThrows())).didThrow()
     ),
-    createTest('twoPromises', async () =>
+    createTest('twoPromises can run in parallel', async () =>
       (
         await it(async () => {
-          let fooResolved: boolean | undefined
-          let barResolvedBeforeFoo: boolean | undefined
-          const foo = async () => {
-            await testObject.wait(2)
-            fooResolved = true
-          }
-          const bar = async () => {
-            await testObject.createArrayBufferAsync()
-            barResolvedBeforeFoo = !fooResolved
-          }
-          await Promise.all([foo(), bar()])
-          return barResolvedBeforeFoo
+          const start = performance.now()
+          // 2s + 2s = ~4s in serial, ~2s in parallel
+          await Promise.all([testObject.wait(2), testObject.wait(2)])
+          const end = performance.now()
+          const didRunInParallel = (end - start) < 4000
+          return didRunInParallel
         })
       )
         .didNotThrow()


mrousavy:
Hm, the `.Default` scope should allow for parallelism - it is a shared pool. https://github.com/mrousavy/nitro/blob/d76d0c02621d4183544d2a8af5b8b334318b4551/packages/react-native-nitro-modules/android/src/main/java/com/margelo/nitro/core/Promise.kt#L57

I'd have to test this more, but in the meantime you can play around with this to see if changing to a different Dispatcher changes things? E.g. a `Executors.newFixedThreadPool(4).asCoroutineDispatcher()`.

mrousavy:
Btw.; didn't you talk about a different issue on WhatsApp? You mentioned that resolve just doesn't work anymore, and that it starts to hang in `JSPromise`, but it worked in `Promise.kt` and `JPromise.hpp`...?

example/src/getTests.ts
```diff
@@ -660,6 +660,26 @@ export function getTests(
...
+          const foo = async () => {
+            await testObject.wait(2)
+            fooResolved = true
+          }
+          const bar = async () => {
+            await testObject.createArrayBufferAsync()
+            barResolvedBeforeFoo = !fooResolved
+          }
+          await Promise.all([foo(), bar()])
+          return barResolvedBeforeFoo
```
mrousavy, line 677: This test is not valid - there is no guarantee that `foo()` finishes before `bar()`, or other way around. That would be a race condition. That's how threads work.

mrousavy (changes requested):
The current test is invalid - we cannot test order of that. It's how native threads work, they can race.

We can test however if both promises resolve one after another - as that was the original issue you sent me on WhatsApp

example/src/getTests.ts
grabbou: Yeah maybe it's invalid - the problem is that the second promise never resolves in the JS, unless first promise is resolved by the native side. On iOS, it works as expected. Promises are resolved on the JS side as soon as they're resolved on the native side. The issue is not about the order per se, but about the fact that bar promise is never resolved, if foo is hanging indefinitely

grabbou:
> Btw.; didn't you talk about a different issue on WhatsApp? You mentioned that resolve just doesn't work anymore, and that it starts to hang in `JSPromise`, but it worked in `Promise.kt` and `JPromise.hpp`...?

Yes, it's still same issue, and this test shows its symptoms! Something stops JS from resolving the promise and my debugger breakpoints reveal that it's not hitting the JSPromise code

grabbou:
> The current test is invalid - we cannot test order of that. It's how native threads work, they can race.
> 
> 
> 
> We can test however if both promises resolve one after another - as that was the original issue you sent me on WhatsApp

Hm, I am not sure if that was the original issue and whether your updated test does show the real problem.

The issue is not promises running in parallel.

The issue is if there are two promises that run in parallel, one is wait 4, second one is wait 2, the second one will not resolve until first one resolved, so eventually they both resolve after 4 seconds

In practice, second one should resolve earlier, and code that follows after its await should be executed while we're waiting for first promise.

In my code, those promises are run from two different places (not Promise.all) but I did that just to demonstrate the case.

I guess I'll have to send a PR with a different test case to make this explicit

example/src/getTests.ts
```diff
@@ -660,6 +660,20 @@ export function getTests(
     createTest('promiseThrows() throws', async () =>
       (await it(() => testObject.promiseThrows())).didThrow()
     ),
+    createTest('twoPromises can run in parallel', async () =>
+      (
+        await it(async () => {
+          const start = performance.now()
+          // 2s + 2s = ~4s in serial, ~2s in parallel
+          await Promise.all([testObject.wait(2), testObject.wait(2)])
```
grabbou, line 668: I think a better test case would be to check how long did each promise take to resolve.

So, make wait 4, wait 2, and make sure first one took 4s, second one 2s to resolve.

What you'll find out, unless your rewrite did fix things, is that they will both take same time to resolve (4s)

example/src/getTests.ts
grabbou: BTW note that both foo and bar had different time to resolve, so it was valid in my opinion to check for the order they're resolved, if one takes 2 seconds and second one is immediate (no coroutine on the native side)

grabbou:
Okay, I run this test on main **and issue is gone. this passes**. Promises resolve after the right time (4 and 1s respectively). Before, it would be always 4. Potentially related https://github.com/mrousavy/nitro/pull/347

```ts
 createTest('twoPromises can run in parallel', async () =>
      (
        await it(async () => {
          const foo = async () => {
            await testObject.wait(4)
            const end = performance.now()
            return end - start < 4 * 1000 + 10
          }
          const bar = async () => {
            await testObject.wait(1)
            const end = performance.now()
            return Math.round(end - start) < 1 * 1000 + 10
          }
          const [fooTime, barTime] = await Promise.all([foo(), bar()])
          return [fooTime, barTime]
        })
      )
        .didNotThrow()
        .equals([true, true])
    ),
```

mrousavy:
Yea I might've fixed this in #347 🙈

PR #341, opened 2024-11-17
Use "nitrogen" as bin

kraenhansen:
Merging this will declare the "bin" of "nitro-codegen" as "nitrogen", allowing any package taking on the "nitro-codegen" to execute the local "nitrogen" binary. I've updated docs and usage to the best of my abilities 🤞

### DIFF
--- a/docs/docs/nitrogen.md
+++ b/docs/docs/nitrogen.md
@@ -119,22 +119,22 @@ Now run nitrogen:
 <Tabs groupId="package-manager">
   <TabItem value="npm" label="npm" default>
     ```sh
-    npm run nitro-codegen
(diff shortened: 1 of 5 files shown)


docs/docs/nitrogen.md
```diff
@@ -119,22 +119,22 @@ Now run nitrogen:
 <Tabs groupId="package-manager">
   <TabItem value="npm" label="npm" default>
     ```sh
-    npm run nitro-codegen
+    npx nitrogen
```
kraenhansen, line 122: This was making assumptions on the existence of a "nitro-codegen" script in the package, which is not way I believe was the intent.

PR #350, opened 2024-11-19
feat: Add `stacktrace` to errors in Nitro

mrousavy:
Adds `stacktrace` to all errors - in Java this comes from `JThrowable`.


mrousavy:
This is lots of effort for very little benefit. I think I'm gonna drop this.

Issue #93, opened 2024-08-31
Nitro 0.6 - Cannot default-construct HybridObject!


jpudysz:
Hey!
After upgrading from v.0.5.0 to v.0.6.0, I can no longer construct my Hybrid that depends on another Cxx Hybrid

```cpp
 auto nativePlatform = Unistyles::NativePlatform::create(); // cxx <-> swift hybrid
 auto unistylesRuntime = std::make_shared<HybridUnistylesRuntime>(nativePlatform); // <- crashes here
```

Error:
```
libc++abi: terminating due to uncaught exception of type std::runtime_error: Cannot default-construct HybridObject!
```

It would be great to still support that, as I don't want to expose `NativePlatform` to JS.

PR #366, opened 2024-11-22
docs: :memo: add explicitly call to HybridObject constructor

DavideSegullo:
Hi, it seems like that [from version 0.6.0](https://github.com/mrousavy/nitro/issues/93) it's required to call `HybridObject` constructor since it's virtual, this PR add this suggestions inside the docs.

Maybe would be also helpful to specify it somewhere in the doc so we can avoid this error at runtime, if someone have any suggestion I can it ofc:

```
Cannot default-construct HybridObject!
```

### DIFF
--- a/docs/docs/nitrogen.md
+++ b/docs/docs/nitrogen.md
@@ -234,6 +234,8 @@ To implement `Math` now, you just need to implement the spec:
         ```cpp title="HybridMath.hpp"
         class HybridMath: public HybridMathSpec {
         public:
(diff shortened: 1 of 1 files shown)


mrousavy:
Good point! I'm trying to keep the docs very simple and easy to read, and I think with this change there is now a layout shift when switching between languages... 😅

DavideSegullo:
Do you see a layout shift because now we have different card height for each language?

PR #374, opened 2024-11-27
fix: Example app ScrollView bottom offset

chrispader:
Adds `paddingBottom` to the style of the `ScrollView` in the example app, so that the last test box is not cut off.

### DIFF
--- a/example/src/screens/HybridObjectTestsScreen.tsx
+++ b/example/src/screens/HybridObjectTestsScreen.tsx
@@ -207,7 +207,9 @@ const styles = StyleSheet.create({
   container: {
     flex: 1,
   },
-  scrollContent: {},
+  scrollContent: {
+    paddingBottom: 65,
+  },
   topControls: {
     marginHorizontal: 15,
     marginBottom: 10,


mrousavy:
screenshots pls.

example/src/screens/HybridObjectTestsScreen.tsx
```diff
@@ -207,7 +207,9 @@ const styles = StyleSheet.create({
   container: {
     flex: 1,
   },
-  scrollContent: {},
+  scrollContent: {
+    paddingBottom: 65,
```
mrousavy, line 211: We don't do magic values. We have `react-native-safe-area-context`, please use it.

mrousavy:
closed in favor of 196f2aef

example/src/screens/HybridObjectTestsScreen.tsx
chrispader: The change isn't because of the safe area, but because of the overlay at the bottom ("Run tests", etc.)

mrousavy: yea i understand - but lets avoid magic numbers and just make it relative not absolute

PR #376, opened 2024-11-29
feat: Allow using `Error` as type

mrousavy:


packages/react-native-nitro-image/ios/HybridTestObjectSwift.swift
```diff
@@ -120,6 +120,11 @@ class HybridTestObjectSwift : HybridTestObjectSwiftKotlinSpec {
     return 55
   }
 
+  func throwError(error: Error) throws -> Void {
+    // TODO: Swift functions can not throw yet! Errors are not propagated up to C++.
+    // throw error
+  }
```
renanmav, line 126: What's the use case if swift functions cannot throw? To read those JS errors from the native side?

mrousavy: Yea. In Swift it's a compiler limitation that we cannot throw to C++.

I _could_ fix it by making every method return a `variant<TResult, Error>` instead of just `TResult`, then I can forward-propagate them to C++. Comes with a performance tradeoff, but is a good temporary solution.
I think there was just something broken with variants before...

PR #379, opened 2024-12-02
fix: Fix build when using `use_frameworks`

mrousavy:
For some reason, Folly seems to spazz out when building the iOS app with `use_frameworks` / `ENV['USE_FRAMEWORKS']` enabled.

Frameworks are a relict of the past anyways and are not really recommended to be used because they're dynamically linked instead of statically compiled modules, which is what Swift supports now. 
But Firebase still seems to use Frameworks and for some reason STILL haven't migrated away from it.

People complained to me that Nitro doesn't build with Firebase, instead of telling Firebase to migrate away from Frameworks.

Anyways.. This now adds support for Frameworks.

For this to work, we have to disable the Folly configuration (`FOLLY_NO_CONFIG`) and coroutines, and import the `-Swift.h` header using the fully qualified name.

I have no idea why Folly seems to fail to build with frameworks, but whatever.

Since this is not really part of Nitro or Nitrogen, users are expected to add this to their `.podspec` files.

This PR is based off of https://github.com/mrousavy/nitro/pull/333 (so huge thanks to the author!), but changed because he didn't format it properly and removed imports and didn't tell me why


baveku:
Bravo, Let's go.... 🚀🚀🚀🚀

Issue #384, opened 2024-12-03
Cannot clone repo on Windows


tido64:
### How were you trying to build the app?

Trying to clone the repo on Windows fails because some files have names containing invalid characters: https://gist.github.com/doctaphred/d01d05291546186941e1b7ddc02034d3

### Full build logs

```tsx
Cloning into 'nitro'...
remote: Enumerating objects: 16397, done.
remote: Counting objects: 100% (2670/2670), done.
remote: Compressing objects: 100% (1123/1123), done.
remote: Total 16397 (delta 2190), reused 1605 (delta 1533), pack-reused 13727 (from 1)
Receiving objects: 100% (16397/16397), 17.85 MiB | 16.55 MiB/s, done.
Resolving deltas: 100% (10960/10960), done.
error: invalid path 'packages/template/<<iosModuleName>>.podspec'
fatal: unable to checkout working tree
warning: Clone succeeded, but checkout failed.
You can inspect what was checked out with 'git status'
and retry with 'git restore --source=HEAD :/'
```

### Project dependencies

```json
n/a
```

### Nitro Modules Version

main

### Nitrogen Version

main

### Target platforms

Android

### Operating system

Windows

### Can you build the Nitro Modules Example app?

No, I cannot build the Example app either

### Additional information

- [ ] I am using Expo
- [ ] I am using Nitrogen (nitro-codegen)
- [X] I have read and followed the [Troubleshooting Guide](https://nitro.margelo.com/docs/troubleshooting).
- [ ] I created a reproduction PR to reproduce this issue here in the nitro repo. (See [Contributing](https://nitro.margelo.com/docs/contributing) for more information)
- [X] I searched for [similar issues in this repository](https://github.com/mrousavy/nitro/issues) and found none.

mrousavy:
Shit I knew this would become an issue. Do you think I can replace `<<something>>` with `[[something]]`?

tido64:
What about `__something__`? Underscore seems safer?

mrousavy:
I'm afraid that underscores might be used somewhere... Python uses `__init__`, not sure if JS, Ruby, Groovy, Swift, C, C++, ObjC, Kotlin, Java or CMake have similar stupid ideas somewhere.

mrousavy:
`$` might work: https://github.com/mrousavy/nitro/pull/385

mrousavy:
Can you try cloning the repo from #385? If that works I'll merge it.

mrousavy:
[...]

PR #385, opened 2024-12-03
fix: Replace placeholders to make them windows compatible (`<<*>>` -> `$$*$$`)

mrousavy:
Fixes https://github.com/mrousavy/nitro/issues/384


atlj:
No worries! We're using ejs templates and it's using something like <%- js-expression -%> to render them

atlj:
@mrousavy I created a custom template derived from the `template/` repo. I've gone through each file in template one by one and made sure we had the crnl counterpart properly done

atlj:
This means changes to the `template/` will have to also be manually done in crnl.

PR #386, opened 2024-12-03
PoC: use raw `jsi::Value` prop w/ fabric views

hannojg:
## How to setup:

(I had to update to RN >= 0.76.2 to fix a build issue where some yoga header could not be found)

1. Clean your existing repo:`git clean -dfX`
2. Install packages: `bun i`
3. Apply the RN patch: `npx patch-package`
4. Build lib: `bun run build`
5. Install pods as usual in the example app

## PRs in react-native to get it work:

- https://github.com/facebook/react-native/pull/48243
- https://github.com/facebook/react-native/pull/48231
- https://github.com/facebook/react-native/pull/48202
- https://github.com/facebook/react-native/pull/48232
- https://github.com/facebook/react-native/pull/48047
- https://github.com/facebook/react-native/pull/48255


example/package.json
```diff
@@ -12,7 +12,7 @@
     "lint": "eslint \"**/*.{js,ts,tsx}\" --fix",
     "lint-ci": "eslint \"**/*.{js,ts,tsx}\" -f @jamesacarr/github-actions",
     "bundle-install": "bundle install",
-    "pods": "cd ios && bundle exec pod install && rm -rf .xcode.env.local",
+    "pods": "cd ios && bundle exec pod install",
```
hannojg, line 15: btw: without this change i can't run the example app. I am pretty sure that deleting the .xcode.env.local is wrong.
Will remove this change in a moment though

hannojg:
Okay, i am still missing one piece, seems like the props parser currently removes our prop:

> [RawPropsParser.cpp:75] Looked up property name which was not seen when preparing: nativeProp

fixing …

packages/react-native-nitro-image/android/CMakeLists.txt
```diff
@@ -21,10 +21,15 @@ include_directories(
...
+find_package(fbjni REQUIRED CONFIG) # <-- Used for communication between Java <-> C++
+find_package(ReactAndroid REQUIRED CONFIG) # <-- Used to set up React Native bindings (e.g. CallInvoker/TurboModule)
 
 # Link all libraries together
 target_link_libraries(
         ${PACKAGE_NAME}
         ${LOG_LIB}
+        fbjni::fbjni                              # <-- Facebook C++ JNI helpers
+        ReactAndroid::jsi                         # <-- RN: JSI
+        ReactAndroid::reactnative
```
mrousavy, line 33: all of those should already exist in the `+autolinking.cmake` file

hannojg: yeah, can be removed here

hannojg:
@mrousavy android is good as well now. 

In the same phase in which the new props are updated we get the update on the java side, so I think there is no weird back and forth.

We have to use the state wrapper on android as this is the only thing thats shared between c++ and java for the view manager (props are always eagerly converted to folly::dynamic).

hannojg:
Wait, updating the patch file …

example/src/components/CustomView.ts
```diff
@@ -0,0 +1,23 @@
...
+  },
+}
+
+// TODO: try passing a nitro object here and casting it on the native side!
+type NativeProps = ViewProps & {
+  nativeProp: TestObjectSwiftKotlin
+}
+
+export const CustomView: HostComponent<NativeProps> =
+  NativeComponentRegistry.get<NativeProps>('CustomView', () => viewConfig)
```
mrousavy, line 23: Why does [`requireNativeComponent(..)`](https://github.com/facebook/react-native/blob/main/packages/react-native/Libraries/ReactNative/requireNativeComponent.d.ts) not work here?

hannojg: it might, we'd have to test this

packages/react-native-nitro-image/ios/NitroExampleView.mm
```diff
@@ -0,0 +1,90 @@
...
+public:
+  CustomViewProps() = default;
+  CustomViewProps(const PropsParserContext& context, const CustomViewProps& sourceProps, const RawProps& rawProps)
+      : ViewProps(context, sourceProps, rawProps) {
+        if (rawProps.isEmpty()) {
+          // There is some code path in RawPropsParser.cpp where we expect to collect all possible keys (prop names) upfront
+          // during the init phase. Calling this .at() method while the rawProps is empty, will cause this prop value to be indexed …
+          // Otherwise we'd get a crash when trying to access it later.
+          // We should really also try to merge the PR i have for providing alternative RawProps parser !!
+          rawProps.at("nativeProp", nullptr, nullptr);
```
mrousavy, line 36: Not sure if I fully understand - is this still needed? Do I need to do this for _every_ prop?

hannojg: but yeah, it has to be called at this point once, as the `.at()` method will insert the props key into a map. If later on the prop name is not in this key map it will always return null. Super weird I know.
my main idea is that we are able to write our own prop parser to work around this

packages/react-native-nitro-image/nitrogen/generated/android/NitroImage+autolinking.cmake
```diff
@@ -66,3 +69,20 @@ else()
...
+        -DFOLLY_HAVE_CLOCK_GETTIME=1
+        -DFOLLY_USE_LIBCPP=1
+        -DFOLLY_CFG_NO_COROUTINES=1
+        -DFOLLY_MOBILE=1
+        -DFOLLY_HAVE_RECVMMSG=1
+        -DFOLLY_HAVE_PTHREAD=1
+        # Once we target android-23 above, we can comment
+        # the following line. NDK uses GNU style stderror_r() after API 23.
+        -DFOLLY_HAVE_XSI_STRERROR_R=1
+)
```
mrousavy, line 88: Do I really need those..? @hannojg

packages/react-native-nitro-image/nitrogen/generated/android/c++/CustomComponentDescriptor.h
```diff
@@ -0,0 +1,114 @@
...
+                                                          ViewEventEmitter, // default one
+                                                          CustomStateData> {
+
+public:
+  ComponentShadowNode(const ShadowNodeFragment& fragment, const ShadowNodeFamily::Shared& family, ShadowNodeTraits traits)
+      : ConcreteViewShadowNode(fragment, family, traits) {}
+
+  ComponentShadowNode(const ShadowNode& sourceShadowNode, const ShadowNodeFragment& fragment)
+      : ConcreteViewShadowNode(sourceShadowNode, fragment) {}
+};
```
mrousavy, line 89: Do I need this for android, or will the iOS code from above (`using ... = ...`) work instead of subclassing it here?

PR #390, opened 2024-12-04
chore(nitro-config): adds a json schema for the nitro config file

patrickkabwe:
Adds a json schema for the nitro config file to improve DX.
image


mrousavy:
yo this is awesome!! i didnt know this was a thing.

nice man. is there any way we can test or validate this in CI?

PR #395, opened 2024-12-05
feat: codegen based on platform

patrickkabwe:
Currently, nitrogen generates code for all platforms (code/hybrid objects) regardless of the platform specified in `<{platform:lang}>` generic tags. This PR modifies the code generation to respect these platform tags and only generate the necessary code for the specified platforms.

For example:
- `<{ios:'swift'|'c++'}>` will only generate iOS-specific and shared code
- `<{android:'kotlin'|'c++'}>` will only generate Android-specific and shared code
- `<{ios:'swift'|'c++', android:'kotlin'|'c++'}>` will generate Android, iOS and shared code

This change improves efficiency by preventing unnecessary code generation for unused platforms.


packages/nitrogen/src/nitrogen.ts
```diff
@@ -101,7 +103,8 @@ export async function runNitrogen({
           continue
         }
 
-        const platforms = Object.keys(platformSpec) as Platform[]
+        platforms = Object.keys(platformSpec) as Platform[]
```
mrousavy, line 106: This overwrites `platforms` each time instead of appending to it.

mrousavy (changes requested):

packages/nitrogen/src/nitrogen.ts
mrousavy: Instead of lifting out `platforms` here, let's create a new value in line 82 called `usedPlatforms` and we additionally append to it.

patrickkabwe: addressed the issue above

patrickkabwe:
Hey @mrousavy, anything else you'd like me to check or tweak on this PR?

mrousavy:
Hey - I'm still thinking about this.

The main reason why I don't want to merge this is because it is not tested in the example here.

There is no module that's iOS or Android only.
I'm afraid that there will be some code in the future that assumes iOS/Android files are there, and it fails on libraries that are iOS/Android only because we didn't test it here in the example.

patrickkabwe:
> Hey - I'm still thinking about this.
> 
> The main reason why I don't want to merge this is because it is not tested in the example here.
> 
> There is no module that's iOS or Android only. I'm afraid that there will be some code in the future that assumes iOS/Android files are there, and it fails on libraries that are iOS/Android only because we didn't test it here in the example.

Alright thanks for the feedback. i will add some tests for this over the weekend.

mrousavy:
dont worry, lets just use it as is.

PR #405, opened 2024-12-06
feat: Add `createConstant(...)` API

mrousavy:
Adds an API that allows users to cache constants on JS level, in the prototype chain.

When calling `createConstant(hybrid, 'property')`, the function accesses `.property` once (calls the native C++ getter once), and then caches it in it's prototype.

The implementation traverses the prototype chain so that it adds it to the exact prototype that defines the property, and throws if it cannot be found.

```ts
const math1 = NitroModules.createHybridObject<Math>('Math')
const math2 = NitroModules.createHybridObject<Math>('Math')
math1.pi // <-- accesses native code
math2.pi // <-- accesses native code
createConstant(math1, 'pi')
math1.pi // <-- is cached on JS level
math2.pi // <-- is cached on JS level
```


packages/react-native-nitro-modules/src/Constants.ts
```diff
@@ -0,0 +1,49 @@
...
+ * const math1 = NitroModules.createHybridObject<Math>('Math')
+ * const math2 = NitroModules.createHybridObject<Math>('Math')
+ * math1.pi // <-- accesses native code
+ * math2.pi // <-- accesses native code
+ * createConstant(math1, 'pi')
+ * math1.pi // <-- is cached on JS level
+ * math2.pi // <-- is cached on JS level
+ * ```
+ */
+export function createConstant<T extends HybridObject>(
```
renanmav, line 22: would it be helpful to add control over this JS vs native caching level?

mrousavy: Well it doesn't make sense to add native caching. We want to avoid a JS -> native call, that's why we cache it in JS. That's what this function does

mrousavy:
idk, this is kinda unnecessary.
Library authors can create constants much easier;

```ts
const Math = NitroModules.createHybridObject<Math>('Math')
export const PI = Math.pi
```

PR #419, opened 2024-12-11
fix: `react` and `react-native` dependency resolutions

chrispader:
Since bun uses the latest package versions for dependencies wildcarded with `*`, we will have to set React specifically, since `docs` use a newer React version.

To be consistent with the recent changes in `example` and to avoid any future problems (with RN as well) i changed the dependencies to specific versions instead of `*`


mrousavy:
thanks!

the problem is that dependabot will still try to upgrade `react` to 19, but I think i cna just ignore that

chrispader:
Ah ok, we can look into upgrading React to v19 in a later PR.

PR #430, opened 2024-12-18
feat: Implement Swift errors 🥳

mrousavy:
Finally implements errors for synchronous Swift functions.

Every Swift function is now wrapped with a `Result<T>` type that is either `T` (`hasValue()`) or `std::exception_ptr` (`hasError()`).

This adds a tiny bit of runtime overhead, but at least now we can throw errors that can actually be caught in JS without crashing the app.

I hope in a future version of Swift, Swift errors can actually be caught in C++ so that we can remove this result type wrapping and instead use normal try/catch flows....

### DIFF
--- a/example/ios/Podfile.lock
+++ b/example/ios/Podfile.lock
@@ -1895,7 +1895,7 @@ SPEC CHECKSUMS:
   glog: 08b301085f15bcbb6ff8632a8ebaf239aae04e6a
   hermes-engine: 06a9c6900587420b90accc394199527c64259db4
   NitroImage: df6e7bc6d5cc2be5d3c47aec9e88d2af8498da88
-  NitroModules: d5e3886c34fc5c5ebab18327e52f35460c039e2d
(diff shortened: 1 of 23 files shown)


example/src/getTests.ts
```diff
@@ -428,6 +428,13 @@ export function getTests(
         `Error: ${testObject.name}.funcThatThrows(...): This function will only work after sacrificing seven lambs!`
       )
     ),
+    createTest('funcThatThrowsBeforePromise() throws', async () =>
+      (
+        await it(async () => await testObject.funcThatThrowsBeforePromise())
+      ).didThrow(
+        `Error: ${testObject.name}.funcThatThrowsBeforePromise(...): This function will only work after sacrificing eight lambs!`
```
renanmav, line 435: lol

grabbou:
Awesome, very excited to try this out! Just upgrading the library to 0.20.0!

PR #435, opened 2024-12-19
perf: Mark all types that are used only in `std::shared_ptr<T>` as `SWIFT_NONCOPYABLE`

mrousavy:
### Summary

This PR avoids copying a few C++ types when using them in Swift, effectively speeding up the codebase by a few factors depending on how heavy the types were to copy. 🔥 

### Problem

Swift imports all C++ types as value types, so every time you get, use, or pass a C++ type around in Swift, **it will make a copy.**

Not only is this slow, but sometimes it isn't even expected and might destroy internal state if the C++ class has unsafe pointers inside that are not ref-counted! ⚠️ ❌ 

`std::shared_ptr<T>` is one of these cases - if you use `shared_ptr.pointee`, it will actually return the pointer this shared_ptr wraps **by value** - so **it will copy** `T`!! ❌ 

```swift
func someSwiftFunc(shared: std.shared_ptr<T>) {
  shared.pointee.doFirst()
  shared.pointee.doSecond()
  shared.pointee.doThird()
  // 3x copy, 3x destructors!! ❌ ❌ ❌ 
}
```

### Solution

So this PR changes this - every type we use inside of a `std::shared_ptr` is now marked as a `SWIFT_NONCOPYABLE`, meaning Swift will not copy this type. It is like `SWIFT_IMMORTAL_REFERENCE`, but a bit more explicit on the usage.

Right now, the following types are affected and have been improved with this PR;

- `Promise<T>`
- `ArrayBuffer`
- Callbacks / `std::function`

### Future

Every type that is not expected to be copied directly should have `SWIFT_NONCOPYABLE` on it, so that's;
- Extrusively reference-counted types (aka every type in a `shared_ptr<T>`, `unique_ptr<T>` or `weak_ptr<T>`)
- `const &` types
- Unsafe pointers

Then Swift can use borrowing magic.

Also, I guess an even better solution would be to have a `borrowing T` accessor instead of just `T` on the Swift shared_ptr type. I created a feature request for this here: https://github.com/swiftlang/swift/issues/78296

### DIFF
--- a/packages/nitrogen/src/syntax/swift/SwiftCxxTypeHelper.ts
+++ b/packages/nitrogen/src/syntax/swift/SwiftCxxTypeHelper.ts
@@ -377,7 +377,7 @@ public:
   }
 private:
   ${actualType} _function;
-} SWIFT_UNSAFE_REFERENCE;
+} SWIFT_NONCOPYABLE;
 inline ${name} create_${name}(void* _Nonnull closureHolder, ${functionPointerParam}, void(* _Nonnull destroy)(void* _Nonnull)) {
   std::shared_ptr<void> sharedClosureHolder(closureHolder, destroy);
   return ${name}([sharedClosureHolder = std::move(sharedClosureHolder), call](${paramsSignature.join(', ')}) -> ${type.returnType.getCode('c++')} {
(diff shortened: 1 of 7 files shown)


mrousavy:
Shit, this requires min iOS version of iOS 16.

mrousavy:
omg `SWIFT_NONCOPYABLE` WORKS!!! 🥳

PR #438, opened 2024-12-19
feat: Add `PromiseHolder<T>` to avoid expensive copies of `Promise<T>`

mrousavy:


mrousavy:
actually I'm not sure if I really want to spend the effort to build this. This is added complexity just for some minor performance gains - and in the future I'm hoping there will be a `borrowing T` accessor for `shared_ptr`, then we don't need this anyways.

PR #433, opened 2024-12-19
test: Add test to call async callback

mrousavy:
Adds a test to call an asynchronous callback

```ts
  callbackAsyncPromise(callback: () => Promise<number>): Promise<number>
```

~~With this change, I can now reproduce the memory corruption bug:~~

```diff
- NitroExample(36309,0x104f4c580) malloc: Heap corruption detected, free list is damaged at 0x281e3f5a0
*** Incorrect guard value: 36979507128897
- NitroExample(36309,0x104f4c580) malloc: *** set a breakpoint in malloc_error_break to debug
```

~~I reported this already at the Swift compiler repo: https://github.com/swiftlang/swift/issues/78292~~

The above crash has finally been fixed in https://github.com/mrousavy/nitro/pull/440!

PR #440, opened 2024-12-20
feat: Create Swift closure wrapper

mrousavy:
Instead of bending C-style function pointer magic, I rewrote the function parsing logic to use statically generated classes instead.

This might result in better performance, but is also much safer and has more stability as the memory is strongly referenced using ARC.

Previously we counted references on an unsafe `void*`.

Additionally, it _might_ fix the `malloc: Heap corruption detected ***` crash reported in https://github.com/mrousavy/nitro/pull/433! 👀


mrousavy:
cc @ChrisSchofieldCheckatrade try this

PR #451, opened 2024-12-27
fix: Use UNIX paths for CMake file, even on Windows (`\` -> `/`)

mrousavy:
Fixes Windows support for Android codegen

### DIFF
--- a/packages/react-native-nitro-modules/package.json
+++ b/packages/react-native-nitro-modules/package.json
@@ -117,7 +117,6 @@
       [
         "typescript",
         {
-          "tsc": "../../node_modules/.bin/tsc",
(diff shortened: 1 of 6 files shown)


packages/react-native-nitro-modules/package.json
```diff
@@ -117,7 +117,6 @@
       [
         "typescript",
         {
-          "tsc": "../../node_modules/.bin/tsc",
```
mrousavy, line 120: Adding it manually here failed on Windows.
Now that we removed it, it still works on Unix because it just recursively goes up to find `tsc`. Same story on Windows now, previously Windows didn't do that because of whatever reason.

PR #468, opened 2025-01-08
Updated the Math interface example in the Nitro Modules documentation.

mensonones:
This PR updates the Math interface example in the Nitro Modules documentation by adding a missing HybridObject import. Without the explicit import, users copying the example could encounter errors, as HybridObject would not be recognized.

Error without import

image


mrousavy:
Thanks for your PR, but I'd like to keep the examples clean - and this would error in your TS project anyways. I think it's a fair assumption that the user understands he needs to import that type.

PR #486, opened 2025-01-12
fix: Replace `HybridObjectSpec` (a protocol) with `HybridObject` (a base class)

mrousavy:
In Swift, `HybridObject` is now a base class instead of a protocol.


mrousavy:
ugh it doesnt work because if `HybridMath` (from MyNitroModule) extends `HybridObject` (from Nitro Core), it cannot resolve the base class because `HybridObject` is from a different clang module than `HybridMath`.

PR #500, opened 2025-01-15
feat: Create React Native ViewManager (+ Props + State) for HybridViews

mrousavy:
Actually update props in the Swift/Kotlin view when they change in C++/JSI.

This makes Nitro Views ready for beta testing


mrousavy:
blocked because of this bs 
image

hannojg:
Ah, there might be some caveat about passing a callback as prop. These are implemented differently in react-native/fabric. They are just hold on the native side and a general purpose event emitter emits and event to js to call this callback and then its called from the JS side (if I am not mistaken).
So it might be that callbacks are handled differently during prop parsing 🥶 will check / confirm alter today!

mrousavy:
> (A quick fix would be to box those, but I know how much you would hate that 😄 )

very much

hannojg:
@mrousavy that might happen on the JS side already, not on the native side

mrousavy:
so worst case we need a PR that customizes/removes this behaviour from the JS side - maybe a third param to the getHostComponent thingy

mrousavy (commented):
looks good nice work @mrousavy

mrousavy:
thanks

PR #503, opened 2025-01-21
chore: include file outside of rootDir

hannojg:


mrousavy:
Went a different route in https://github.com/mrousavy/nitro/pull/512. I have JSON now

PR #510, opened 2025-01-22
feat: Update `memorySize` when a HybridObject makes a round-trip

mrousavy:
This internally updates the memory-size each time a HybridObject gets passed goes from native to JS (because it calls `HybridObject::toObject(...)`, which now does that).

Also, a new API is added: `NitroModules.updateMemorySize(hybridObject)` which is doing exactly that.


mrousavy:
cc @KrzysztofMoch

PR #513, opened 2025-01-25
feat: add support for `*.nitro.tsx` extension

### LINKED ISSUE #507: Add support for `*.nitro.tsx` files
Support the .tsx extension for files as well.
If you create a `.nitro.tsx` file nitrogen will complain:

```
Nitrogen didn't find any spec files in ./**/*.nitro.ts! To create a Nitro Module, create a TypeScript file with the ".nitro.ts" suffix and export an interface that extends HybridObject<T>.
```

patrickkabwe:
This change adds support for `.tsx` files in the project.

Closes #507


mrousavy:
Hey patrick -thanks for your PR, but unfortunately I don't think this is a good idea. Nitro specs are not JSX, they are simply TypeScript files.

I don't think .tsx should be supported, this is unnecessary complexity imo.

Thanks anyways for your contributions!

PR #516, opened 2025-02-02
fix: Record fromCpp jni can't find kotlin data class constructor

SamuelScheit:
#### Context
I'm currently working on [react-native-fast-io](https://github.com/callstackincubator/react-native-fast-io) on a spec compliant fetch implementation.
For that the library has to accept a headers argument as `Record<string, string>` type.

### Problem
When specifying the following nitro type
```ts
export type RequestOptions = {
  headers: Record<string, string>
}

export interface Network extends HybridObject<{ ios: 'swift'; android: 'kotlin' }> {
  request(opts: RequestOptions): Promise<void>
}
```

It will generate the following kotlin data class definition
```kotlin
/**
 * Represents the JavaScript object/struct "RequestOptions".
 */
@DoNotStrip
@Keep
data class RequestOptions
  @DoNotStrip
  @Keep
  constructor(
    val headers: HashMap<String, String>
  ) {
  /* main constructor */
}
```

and the following c++ JRequestOptions fromCpp definition:
```c++

  public:
    /**
     * Create a Java/Kotlin-based struct by copying all values from the given C++ struct to Java.
     */
    [[maybe_unused]]
    static jni::local_ref<JRequestOptions::javaobject> fromCpp(const RequestOptions& value) {
      return newInstance(
        [&]() {
          auto __map = jni::JHashMap<jni::JString, jni::JString>::create(value.headers.size());
          for (const auto& __entry : value.headers) {
            __map->put(jni::make_jstring(__entry.first), jni::make_jstring(__entry.second));
          }
          return __map;
        }()
      );
    }
  };

```
which will create a `jni::JHashMap` and then call `newInstance` with the following type signature `Ljava/util/HashMap;`
which will then fail because it can't find the data class constructor,
because the requested native type is `java.util.HashMap` but the kotlin constructor accepts any `kotlin.collection.Map`.

To fix this the kotlin data class type needs to be changed to HashMap.

I've tested the change and it fixed the problem. Tho I don't know if this will break at other points and the change needs a proper review.

### DIFF
--- a/packages/nitrogen/src/syntax/types/RecordType.ts
+++ b/packages/nitrogen/src/syntax/types/RecordType.ts
@@ -29,7 +29,7 @@ export class RecordType implements Type {
       case 'swift':
         return `Dictionary<${keyCode}, ${valueCode}>`
       case 'kotlin':
-        return `Map<${keyCode}, ${valueCode}>`
+        return `HashMap<${keyCode}, ${valueCode}>`
       default:
         throw new Error(
           `Language ${language} is not yet supported for RecordType!`


mrousavy:
Also I'm seeing now that we don't even have a test for `Record<...>` in the `TestObject.nitro.ts`.

We should add this to make sure Maps work - both for Swift/Kotlin and for C++ (e.g. via a `bounceMap(map: Record<string, T>): Record<string, T>` in `SharedTestObjectProps`)

Can you add this or do you want me to add this?

SamuelScheit:
When invoking the Network.request method it takes the JS args and tries to convert it to RequestOptions type by calling `JRequestOptions::fromCpp`

image

`JRequestOptions` converts the js args into jni/kotlin compatible types.
The js `Record<string,string>` types gets converted to a`jni::JHashMap`.
The resulting args are passed to newInstance.

image

It then tries to get the constructor for the args

image

It converts the passed args to the following jni type  `(Ljava/lang/String;Lcom/margelo/nitro/fastio/RequestMethod;Lcom/margelo/nitro/fastio/HybridInputStreamSpec;Ljava/util/HashMap;)V` 

image

However it fails to find the method with `method=NULL` because the RequestOptions doesn't have a constructor for HashMap, it only accepts generic maps.

```kotlin
data class RequestOptions
  @DoNotStrip
  @Keep
  constructor(
    val url: String,
    val method: RequestMethod,
    val body: HybridInputStreamSpec?,
    val headers: Map<String, String> // here is map used
  ) {
  /* main constructor */
}
```

It would be nice if you could add a test for this

SamuelScheit:
I've just tried and the problem can also be fixed by changing the return type of this closure:
https://github.com/mrousavy/nitro/blob/fe3e22da35b9b41e9d9d1e9900d517bbd73ccf6f/packages/nitrogen/src/syntax/kotlin/KotlinCxxBridgedType.ts#L505-L511

To this type `[&]() -> jni::local_ref<jni::JMap<jni::JString, jni::JString>> {` 

I'll let you decide whats the better fix

mrousavy:
Added a simple test for `Record` to the example module: https://github.com/mrousavy/nitro/pull/517

SamuelScheit:
No problem, thank you very much for addressing the issue

mrousavy:
Released in [Nitro 0.22](https://github.com/mrousavy/nitro/releases/tag/v0.22.0)! 🥳

PR #536, opened 2025-02-12
fix: crash when no modules are registered

tido64:
`getAllHybridObjectNames()` might return an empty array but we're trying to access the first element without checking the size.


mrousavy:
Thanks! Good catch

PR #538, opened 2025-02-12
perf: Cache last `getExternalMemorySize()` to avoid Runtime calls each time

mrousavy:
so this is a micro-optimization. not sure if we want to keep it because it makes the code a tiny bit more complex.

When a C++ `HybridObject` gets passed to JS, `toObject(runtime)` is called on it. This always updates the `setExternalMemoryPressure(..)` on the object so the JS GC knows it's weight.

Previously we always set it, even if the memory pressure is the same size. Now we compare it to it's previous value before re-setting it again.


mrousavy:
> Previously we always set it, even if the memory pressure is the same size. Now we compare it to it's previous value before re-setting it again.

This kinda begs the question - shouldn't this be compared internally in JSI's `setExternalMemoryPressure` function, which returns early if it is the same value? @tmikov

tmikov:
JSI doesn't have a place where to cache this value.

mrousavy:
Closing because this path is hit very rarely and the performance improvements of this PR aren't worth the added complexity.

PR #559, opened 2025-02-22
docs: :memo: update view components examples to avoid codegen error

DavideSegullo:
I was playing a little bit with new `HybridView` feature, but if I follow the tutorial and use the current example code, I'll get this error when i run `bun nitro-codegen`:

```bash
 ❌  Failed to generate spec for CameraView! Error: Props cannot be null! CameraView<...> (HybridView) requires type arguments.
 ❌  No specs found in Camera.nitro.ts!
```

This PR just updated the example code to extends correct types to avoid this kind of error


mrousavy:
Nice catch thank you!

PR #560, opened 2025-02-22
feat: Create `Callback<R(Args...)>` for more powerful callbacks

mrousavy:
This should be faster for JS Callbacks, and will allow sync callbacks.

Before:

```cpp
void doSomething(std::function<void(double)> callback, std::function<double(double)> callbackAndReturn) {
  // ...
  callback(55.3); // <-- returns void

  callbackAndReturn(55.3); // returns Promise<double>
}
```

After:

```cpp
void doSomething(Callback<void(double)> callback, Callback<double(double)> callbackAndReturn) {
  // ...
  callback(55.3); // <-- returns void as before
  callback.callAsync(55.3); // <-- return Promise<void> to allow awaiting
  callback.callSync(55.3); // <-- returns void as before, BUT is called synchronously/blocking.

  callbackAndReturn(55.3); // <-- returns Promise<double> as before
  callbackAndReturn.callAsync(55.3); // <-- return Promise<double> as before
  callbackAndReturn.callSync(55.3); // <-- returns double because it's called synchronously/blocking
}
```

So the key differences are;

- Callbacks that return `void` can now also be awaited
- Callbacks that return any value or `void` can be called synchronously to avoid any Thread hops or delays. (must be on the same thread)

## Why is this not merged

It works pretty well in C++ because of the power of C++ templates, but there is no good way to represent these callbacks in Swift yet. There are no variadic type parameters, and Swift Lambdas cannot be subclassed. Any ideas welcome, but I don't think this can be represented in Swift unless I generate hardcoded classes for each type specialization (e.g. `Callback<void(double)>` -> `Callback_void_double`) - but this is ugly syntax for the user.

### DIFF
--- a/example/ios/Podfile.lock
+++ b/example/ios/Podfile.lock
@@ -1921,7 +1921,7 @@ SPEC CHECKSUMS:
   glog: eb93e2f488219332457c3c4eafd2738ddc7e80b8
   hermes-engine: b417d2b2aee3b89b58e63e23a51e02be91dc876d
   NitroImage: e6ea33d1f353925aac506ace6552e3a8a39a078a
-  NitroModules: 63a0be80b244c1bffbc2d22761671ced35664d79
+  NitroModules: c67a3685429aa915f27df4b351bc619b96e8d64b
   RCT-Folly: 36fe2295e44b10d831836cc0d1daec5f8abcf809
   RCTDeprecation: b2eecf2d60216df56bc5e6be5f063826d3c1ee35
   RCTRequired: 78522de7dc73b81f3ed7890d145fa341f5bb32ea
--- a/packages/react-native-nitro-modules/NitroModules.podspec
+++ b/packages/react-native-nitro-modules/NitroModules.podspec
@@ -33,6 +33,7 @@ Pod::Spec.new do |s|
     # Public C++ headers will be exposed in modulemap (for Swift)
     "cpp/core/AnyMap.hpp",
     "cpp/core/ArrayBuffer.hpp",
+    "cpp/core/Callback.hpp",
     "cpp/core/HybridObject.hpp",
     "cpp/core/Promise.hpp",
     "cpp/entrypoint/HybridNitroModulesProxy.hpp",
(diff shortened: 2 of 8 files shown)


mrousavy:
Okay so note for future self in case I forget this again;

- `Callback<R(Args...)>` is a cool idea in C++
    - `runAsync()` returns a `Promise<T>`, even if `T` is `void` - this is quite efficient and can await the full completion - until the JS callback has actually finished executing - from any arbitrary thread.
    - `runSync()` returns `T` - and runs the JS callback fully synchronously, just as if it was any other C++ function. Pretty cool, and allows us to implement Worklets with Nitro. Thread-checking only happens in Debug for performance reasons. 
    - `operator()` would be async by default, unless the type is a non-void, non-`Promise<*>` type - then it can be sync. can Ce `constexpr`, with no runtime overehad.
- `Callback<R(Args...)>` would also be used in `Promise` (`addOnResolvedListener(...)`) because of how things work.
- `JSIConverter+Callback` would require a few forward-declarations, nothing too impossible.
- `Callback<R(Args...)>` in C++ can be implemented in one of two ways: 
    1. Make `Callback<R(Args...)>` an abstract class with two sub-classes as implementations: `JSCallback` and `NativeCallback` - this requires any parameter of `Callback` to be wrapped in a `shared_ptr` to avoid object-slicing when copying the base class.
    2. Make `Callback<R(Args...)>` a single class with a `union` that splits between JS-, and Native- callback. More efficient, and doesn't require heap allocations! Bit complex, but I implemented it in https://github.com/mrousavy/nitro/pull/561.
- `Callback<R(Args...)>` _could_ be implemented in Kotlin since lambdas are just objects that can be extended. I can add a `callSync(...)` method as an extension to the same type that does a downcast.
- `Callback<R(Args...)>` can **NOT** be implemented in Swift, since Swift does not have variadic generic arguments for protocols, and Swift lambdas cannot be subclassed. So the only way to implement this would be to generate hardcoded types like `Func_void_int` (as opposed to `(Int) -> Void`) which is imo way too ugly. <-- **BLOCKER ❌**

mrousavy:
Closing in favor of https://github.com/mrousavy/nitro/pull/564 - which now finally works. Two separate types, fast, perfect-forwarding, and Swift/Kotlin compatible.

There are two (unavoidable) downsides with the approach though:
1. Whether a callback will be called synchronously or asynchronously has to be decided upfront in the TypeScript spec. You cannot choose to either call it sync or async inside your native method later on - you have to know upfront when defining the specs.
2. A callback that returns `void` is ambiguous between sync and async - so it is async (shoot and forget) by default. If you need a sync callback that doesn't return anything, just return a dummy value like a `boolean`.

PR #562, opened 2025-02-22
docs: fix typo in nitro views documentation

mfkrause:
Stumbled across a small typo (`isBlue` instead of `enableFlash` in the Swift example) in the documentation while I was checking out Nitro Views (great work btw). Confused me for a second, so fixed it. :)


mrousavy:
thanks so much! I must've been drunk when i wrote this

Issue #531, opened 2025-02-10
JNI can't find some Kotlin functions for SDK 24-26


jpudysz:
### What's happening?

Hey,

for older Android SDKs 24-26 JNI struggles to get class method. It throws error that there is no such method, which is obviously not true. It works fine for SDK greater than 26+.

I was able to reproduce it locally, but can't with Nitro example.
I use C++ bindings to store C++ lambda on Kotlin side. I'm using it for sync callbacks.

Removing `private` from `registerPlatformListener_cxx` fixes the issue.

### Reproduceable Code

```tsx
void JHybridNativePlatformSpec::registerPlatformListener(const std::function<void(const std::vector<UnistyleDependency>& /* dependencies */, const UnistylesNativeMiniRuntime& /* miniRuntime */)>& callback) {
    static const auto method = _javaPart->getClass()->getMethod<void(jni::alias_ref<JFunc_void_std__vector_UnistyleDependency__UnistylesNativeMiniRuntime::javaobject> /* callback */)>("registerPlatformListener_cxx"); // <- crashes here
    method(_javaPart, JFunc_void_std__vector_UnistyleDependency__UnistylesNativeMiniRuntime_cxx::fromCpp(callback));
  }
```

### Relevant log output

```shell
Exception in HostObject::get for prop 'Unistyles': java.lang.NoSuchMethodError: no non-static method "Lcom/unistyles/NativePlatformAndroid;.registerPlatformListener_cxx
```

### Device

Android Emulator with SDK 24-26

### Nitro Modules Version

0.22.1

### Nitrogen Version

0.22.1

### Can you reproduce this issue in the Nitro Example app here?

No, I cannot reproduce the issue in the Example app

### Additional information

- [x] I am using Expo
- [x] I am using Nitrogen (nitro-codegen)
- [x] I have read and followed the [Troubleshooting Guide](https://nitro.margelo.com/docs/troubleshooting).
- [ ] I created a reproduction PR to reproduce this issue here in the nitro repo. (See [Contributing](https://nitro.margelo.com/docs/contributing) for more information)
- [x] I searched for [similar issues in this repository](https://github.com/mrousavy/nitro/issues) and found none.

mrousavy:
Hm, I wouldn't understand why this suddenly fails. You can call private methods through JNI, that shouldn't make a difference at all.
[...]

PR #582, opened 2025-02-24
fix: Replace `getClass()` with `javaClassStatic()`

mrousavy:
Replace JNI `getClass()` calls with the more efficient `javaClassStatic()` counterpart.

Instead of going into JNI to look up the class dynamically every time, this just uses the cached static `javaClassStatic()` reference.

Benefits:

- This is faster. All native method calls, and callback calls are now faster!
- This fixes issues like https://github.com/mrousavy/nitro/issues/531 where methods were cached on the wrong Java instance (`getClass()` returned a subclass instead of the defined `javaClassStatic()`)


mrousavy:
fix: Run nitrogen

PR #586, opened 2025-02-25
fix: removes deprecated `HybridObjectRegistry` from template

patrickkabwe:
Fix Android build by removing `HybridObjectRegistry` from the template, following its removal from core in #579.


mrousavy:
ah whoops i forgot about the template, thanks!

PR #603, opened 2025-03-06
feat: Implement `VariantX` for Swift & Kotlin

mrousavy:
Uses a shared indexed Variant type instead of creating new types each time.


mrousavy:
actually https://github.com/mrousavy/nitro/pull/605 is better because we fully avoid boxing and flatten out each type at compile-time.

This is a performance improvement especially for primitives like Double. No need for boxing into `Object` (thats how generics work).

PR #612, opened 2025-03-18
fix: revert Object.freeze

jpudysz:
This PR reverts #602 as it breaks Unistyles.

Freezing objects is indeed a good approach, but it compromises Nitro's flexibility to override certain JS-only properties.  
One use case is mapping React Native `StyleSheet` properties like `compose` and `flatten` to my Hybrid.

I considered exposing some setters on my side, but that feels like overkill for such a simple case.

### DIFF
--- a/packages/react-native-nitro-modules/cpp/core/HybridObject.cpp
+++ b/packages/react-native-nitro-modules/cpp/core/HybridObject.cpp
@@ -79,17 +79,11 @@ jsi::Value HybridObject::toObject(jsi::Runtime& runtime) {
   object.setProperty(runtime, "__type", jsi::String::createFromUtf8(runtime, "NativeState<" + std::string(_name) + ">"));
 #endif
 
-#ifdef NITRO_DEBUG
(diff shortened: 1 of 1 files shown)


mrousavy:
Released in 0.25.2 (from Japan) - :)

jpudysz:
Haha thanks!

jpudysz:
I can confirm it works without any issue!

mrousavy:
Hey JFYI; Prototypes are now frozen - but the individual HybridObjects are not frozen. (starting from Nitro 0.31.0) - I think this is fine, right?

jpudysz:
Yeah, I started getting reports from users about that. It crashes Unistyles. I will try to fix it with new spec generation

mrousavy:
Oh really? 
I mean if it crashes I can revert that again, I probably should've asked you first.

There's no big reason why I would freeze the prototypes, I was just trying to protect users from doing something stupid.

jpudysz:
I will try to figure something out later today and let you know

PR #618, opened 2025-03-22
fix: Don't wrap inherited callbacks in Views (`{ f: … }`)

### LINKED ISSUE #616: [Nitro Views] Inherited callbacks from ViewPropTypes should not be wrapped
### What's happening?

On Nitro Views, Typescript tells us to wrap inherited callbacks from ViewPropTypes (`onLayout`, `onTouchStart`), but doing so will lead to the following error.

We actually do not need to wrap the callback in an object, it will work.

### Reproduceable Code

```tsx
onLayout={{
  f: event => {
    const {width, height} = event.nativeEvent.layout;
    setViewDimensions({width, height});
  },
}}
```

### Relevant log output

```shell
Warning: Error: Exception in HostFunction: Value is an object, expected a boolean
```

### Device

Android

### Nitro Modules Version

0.25.2

### Nitrogen Version

0.25.2

### Additional information

- [ ] I am using Expo
- [x] I am using Nitrogen (nitro-codegen)

patrickkabwe:
may be resolves #616 

Before fix
image

After fix
image


mrousavy:
very good thank you patrick!!

mrousavy:
1. TypeScript lint failed in the CI - pls check that out (I am currently on mobile, in China - so i can't check that rn)
2. To ensure this never breaks again, could you just add a use for that into our ViewsScreen.tsx - eg by adding `onLayout` to the view and just logging it's width + height? Or maybe better the onTouchesDown or whatever its called so it doesn't spam logbox?

patrickkabwe:
> 1. TypeScript lint failed in the CI - pls check that out (I am currently on mobile, in China - so i can't check that rn)
> 2. To ensure this never breaks again, could you just add a use for that into our ViewsScreen.tsx - eg by adding `onLayout` to the view and just logging it's width + height? Or maybe better the onTouchesDown or whatever its called so it doesn't spam logbox?

Done

mrousavy:
letsgo! thanks patrick

Issue #654, opened 2025-05-03
Returning a function in Kotlin causes a `jni::local` crash


patrickkabwe:
### What's happening?

When I define a method that returns a function, like this:

```ts
import { type HybridObject } from 'react-native-nitro-modules'
import type { ConnectionType, NetworkInfoListener } from '../types'

export interface NitroNetworkInfo extends HybridObject<{ ios: 'swift', android: 'kotlin' }> {
    readonly isConnected: boolean
    readonly connectionType: ConnectionType
    addListener(listener: NetworkInfoListener): () => void
}
```
The app crashes on reload (especially hot reload or when calling addListener() again).

I believe the generated native C++ binding for `addListener()` uses a captured local JNI reference `(__result)` in a lambda returned from the method. This reference becomes invalid after the JNI call ends, especially across reloads or GC, leading to undefined behavior or crashes.

```cpp
return [&]() -> std::function<void()> {
  if (__result->isInstanceOf(JFunc_void_cxx::javaClassStatic())) [[likely]] {
    return jni::static_ref_cast<JFunc_void_cxx::javaobject>(__result)->cthis()->getFunction();
  } else {
    return [__result]() {
      __result->invoke();  // ❌ Unsafe: __result is a local_ref, may be invalid
    };
  }
}();
```
`__result` is a `local_ref` valid only during the JNI method scope.
After the method returns, this lambda captures a dangling reference.

On re-invocation (e.g., after a JS reload), the app crashes `(segfault or JVM crash)`.

### Reproduceable Code

```tsx
import { type HybridObject } from 'react-native-nitro-modules'
import type { ConnectionType, NetworkInfoListener } from '../types'

export interface NitroNetworkInfo extends HybridObject<{ ios: 'swift', android: 'kotlin' }> {
    readonly isConnected: boolean
    readonly connectionType: ConnectionType
    addListener(listener: NetworkInfoListener): () => void
}

1. run nitrogen
2. build the app
3. reload app
```

### Relevant log output

```shell
A  java_vm_ext.cc:598] JNI DETECTED ERROR IN APPLICATION: JNI ERROR (app bug): jobject is an invalid local reference: 0x7c1a3ff015 (popped reference at index 5 in a table of size 0)
                                                                                                    java_vm_ext.cc:598]     in call to GetObjectRefType
                                                                                                    java_vm_ext.cc:598]     from void com.facebook.jni.NativeRunnable.run()
[...]

PR #655, opened 2025-05-03
fix(android): prevent crash when returning callback functions by safely capturing weak JNI references

patrickkabwe:
This commit updates the generated C++ glue code to use `jni::make_global()` to avoid invalid memory access and ensure safe interop with Java/Kotlin-managed objects.

Closes: #654


mrousavy:
Hey Patrick - thanks so much for this PR, this is good code.

I have two things, one general approach concern and one code-style concern;


1. I'm not sure if we want to weakify the result here - if it's a `jni::local_ref` it will always be local only, if we weakify it here it will be deleted instantly, and the callback just **never** works since it can never be locked to a strong local again. Don't we want to do a `jni::make_global` here??? 
2. I'd keep the `else` branch to keep it very expressive and easy for the eye that this is the "other" branch.

mrousavy:
Yea I think we want to make it a global, then capture it and then invoke that.

Making it weak will probably cause it to never be called

PR #660, opened 2025-05-09
chore(example): upgrade example app to RN 0.78.2

hannojg:
This fixes a bunch of bundle & runtime errors i was getting on android.

I also had it in other projects that were using RN 0.78.0 that I wasn't able to bundle JS. This fixes it


hannojg:
Can we kick off a CI android build here @mrousavy ?

PR #682, opened 2025-06-05
docs: Update docusaurus

mrousavy:


docs/docusaurus.config.ts
```diff
@@ -29,10 +29,6 @@ const config: Config = {
     locales: ['en'],
   },
 
-  future: {
-    experimental_faster: true,
-  },
```
slorber, line 34: 🥲 why

mrousavy: It broke the build 😭

mrousavy: https://github.com/mrousavy/nitro/pull/694

PR #688, opened 2025-06-06
feat: Add `AnyMap::getAllKeys()` function

### LINKED ISSUE #687: Feature request: Be able to extract all data in AnyMap on the native side (feature parity)
Currently there is no way to extract all the information in AnyMap/AnyMapHolder, only getting individual keys. There is not way to find out what unknown keys might be there.

For example having a `getAllKeys()` method would make it possible to access and extract all the data, even when not knowing the structure beforehand.

(Verified that it's not currently possible in Swift, not in C++ or Kotlin.)

mrousavy:
Adds a `getAllKeys()` function so you know what keys the `AnyMap` contains and can do some looping/iterating.

- Closes #687


mrousavy:
nice, all done;

image

robertherber:
Nice one, thanks! 🙏

PR #694, opened 2025-06-11
docs: Enable faster & v4

mrousavy:
sebastien was about to punch me for disabling it, so lets enable it again!


slorber:
😄 thanks

PR #717, opened 2025-07-08
Potential implementation for adding Swift6 Sendable support to callbacks generated by nitro

TimGass:
I hope that this is the right way to do this, if not feel free to close this pull request. I imagine that this would require documentation changes. I added an optional usesSwift6 option to the ios property of nitro.json because I found that I needed to add ```@preconcurrency``` attributes and ```@Sendable``` attributes to function callbacks generated by nitrogen to get them to not error for non-sendable type in Sendable closures in Swift 6. Admittedly, I'd really like to have this feature in nitrogen because it is the only way that I can build the callbacks inside of Swift 6 sendable closures right now with nitrogen and it would be ideal to not need to manually link everything when using Swift 6. However, I have never made a pull request before, so if I did something wrong or if there is something you would need me to do to make this actually work, feel free to let me know. It has fairly minimal changes. I was just hoping that this code could be looked at and considered. I am not much of a Swift developer, so feel free to let me know if this is not a good implementation and I was not entirely sure if this is the way you would want to implement the changes, but I figured it could be a good template, if not. If there is a more comprehensive way of handling Sendable closures, we can definitely close the pull request, but I could not see much of a way to handle the bridge.wrap_Func calls outside of simply adding the @preconcurrency tag to the hybridObjectSpec method, because I do not know how one would make the C++ Swift Bridged functions comport to the Sendable protocol. It seems, good enough, for now, but likely a patch at best until these things can be sorted out at a later time. Truth be told, I do not entirely know if React Native would even support such a thing at this time. So, this was the best way that I could find in the meantime to solve this scenario. Look forward to your feedback! Thanks for this library it is great!


mrousavy:
Hi - this is an interesting topic. I think I would want to avoid unnecessary extra configuration steps, as this would effectively 2x the test surface of the iOS compiler/generator since we have different output based on this new flag.

So why not just always add `@Sendable`? The lambdas are built to be called from separate Threads... Unless if they are Sync<...>.

I'd have to read into Swift 6 to understand their concepts a bit better tho

mrousavy:
Can you show me the exact error you get when trying to compile as-is on Swift 6?

TimGass:
The problem I was running into in particular was when using a callback with grpc-swift-2 the closures for handling responses are Sendable and callbacks break as calling non-sendables in a sendable closure.

TimGass:
32 |         try await greeter.serverStreamingEvent(SwiftProtobuf.Google_Protobuf_Empty()) { response in
  33 |           for try await event in response.messages {
> 34 |             onEvent(EventMessage(message: event.message, sender: event.sender, sent: Date(timeIntervalSince1970: TimeInterval(event.sent.seconds) + TimeInterval(event.sent.nanos/1000000000))))
     |             ^ capture of 'onEvent' with non-sendable type '-> Void' (aka '(margelo.nitro.grpc.EventMessage) -> ()') in a '@Sendable' closure
  35 |           }
  36 |         }
  37 |       }

TimGass:
My understanding, and you can correct me if I am wrong, is that Swift 6 introduced some new requirements in their compiler for concurrency situations with eliminating race conditions with the Sendable protocol. I do not know what it does, really, but it seems to type color all functions and closures that descend from it, making this a real problem given that I believe Sendable is going to be used everywhere in Swift code that even hints of multithreading after Swift 6. Sadly, this breaks support for all other types, as you can tell, even in the Void return case. I tried simply implementing Sendables, but got stuck on a line like this:
```
      let __result = try self.__implementation.ServerStreaming(onEvent: { () -> (EventMessage) -> Void in
        let __wrappedFunction = bridge.wrap_Func_void_EventMessage(onEvent)
        return { (__event: EventMessage) -> Void in
          __wrappedFunction.call(__event)
        }
      }())
```
I could not find a way to add Sendable to the __wrappedFunction and therefore, decided to add the @preconcurrency attribute to the hybridobjectspec to tell the compiler to ignore it as I do not think there is any way of opting out of this error after Swift 6 besides the @preconcurrency attribute. Swift seems to think that Sendable is a must in multithreaded environments now to avoid race conditions, though, I do not exactly know what it does. It does however seem to enforce it as a requirement on all nested closures, sadly. This is why I thought @preconcurrency was probably the best option at this time.

TimGass:
This is the hybrid object method that was killed by the compiler error.
```
  @available(macOS 15.0, iOS 18.0, watchOS 11.0, tvOS 18.0, visionOS 2.0, *)
  public func ServerStreaming(onEvent: @escaping (_ event: EventMessage) -> Void) throws -> Promise<Void> {
    return Promise.async {
      try await withGRPCClient(
        transport: .http2NIOPosix(
          target: .dns(host: "10.0.0.89", port: 50051),
          transportSecurity: .plaintext
        )
      ) { client in
        let greeter = Greeter.Client(wrapping: client)
        try await greeter.serverStreamingEvent(SwiftProtobuf.Google_Protobuf_Empty()) { response in 
          for try await event in response.messages {
            onEvent(EventMessage(message: event.message, sender: event.sender, sent: Date(timeIntervalSince1970: TimeInterval(event.sent.seconds) + TimeInterval(event.sent.nanos/1000000000))))
          }
        }
      }
    }
  }
  ```

**Comment by TimGass** [comment]
onEvent is a standard nitro callback with its own message type as a parameter for handling callbacks in javascript. It returns Void, but apparently needs to return a Sendable Void with Swift 6's new concurrency model. The only option that I have found to fix this is to add the @Sendable attribute to it and the other callback methods in case they are called from a Sendable closure. The only way I could achieve this with the bridged functions was to disable the compiler warning with @preconcurrency attributes on callback methods, though.

**Comment by TimGass** [comment]
I admit this may not be the optimal solution, but we need some kind of a way of making the callbacks Sendable compatible in Swift 6. Admittedly, I may be overthinking this issue, but I will admit that I am not particularly knowledgeable about how to "break" out of a Sendable closure. So far as I know, this is mainly only a problem for callbacks, which could be run in Sendable closures.

**Comment by mrousavy** [comment]
wdyt about this? https://github.com/mrousavy/nitro/pull/718

**Comment by mrousavy** [comment]
> 32 | try await greeter.serverStreamingEvent(SwiftProtobuf.Google_Protobuf_Empty()) { response in 33 | for try await event in response.messages {
> 
> > 34 |             onEvent(EventMessage(message: event.message, sender: event.sender, sent: Date(timeIntervalSince1970: TimeInterval(event.sent.seconds) + TimeInterval(event.sent.nanos/1000000000))))
> > |             ^ capture of 'onEvent' with non-sendable type '-> Void' (aka '(margelo.nitro.grpc.EventMessage) -> ()') in a '@sendable' closure
> > 35 |           }
> > 36 |         }
> > 37 |       }

This is not nitro related though, is it? This is your code, there's nothing nitrogen should be doing differently here - should it?

**Comment by TimGass** [comment]
This may introduce problems depending on what version of Swift nitro modules wants to support. I am not sure when, but I don't think all versions of Swift have support for @Sendable and @preconcurrency.

**Comment by mrousavy** [comment]
`@Sendable` was introduced in iOS 8. It's a pretty old concept. This will work on all Swift versions.

**Comment by TimGass** [comment]
> `@Sendable` was introduced in iOS 8. It's a pretty old concept. This will work on all Swift versions.

Sounds good to me, then. I just have no idea how you could bridge a C++ function to the Sendable protocol, but if you can, that would be preferable. That was kind of where I got stuck, to be honest.

**Comment by TimGass** [comment]
> > 32 | try await greeter.serverStreamingEvent(SwiftProtobuf.Google_Protobuf_Empty()) { response in 33 | for try await event in response.messages {
> > > 34 |             onEvent(EventMessage(message: event.message, sender: event.sender, sent: Date(timeIntervalSince1970: TimeInterval(event.sent.seconds) + TimeInterval(event.sent.nanos/1000000000))))
> > > |             ^ capture of 'onEvent' with non-sendable type '-> Void' (aka '(margelo.nitro.grpc.EventMessage) -> ()') in a '@sendable' closure
> > > 35 |           }
> > > 36 |         }
> > > 37 |       }
> 
> This is not nitro related though, is it? This is your code, there's nothing nitrogen should be doing differently here - should it?

This would be my hybrid object. The problem being that I cannot run a callback with nitrogen in the HybridObject method's function, if it relies on a Sendable closure, which sadly, in this case, it does. This isn't necessarily a problem with my code or nitrogen's code. The problem is when you call a callback in a context that it doesn't support. For everything else, Sendable does not seem to be an issue, because the different types of code are self-contained. It's just with Swift 6's new concurrency model, not having @Sendable support will make it very hard to interop with most Swift code post Swift 6 and will really hurt the usability of callbacks in nitro modules on that platform. I honestly don't really get why this was added as a mandatory feature in Swift 6, but I am not a Swift guy, admittedly. I just hope it is worth all the pain of updating all the code, in the end.

**Comment by TimGass** [comment]
@mrousavy Proposed changes in that pull request look pretty good to me.

**Comment by mrousavy** [comment]
This is the part that made me worry in my new PR:

https://github.com/mrousavy/nitro/blob/347ce3578f8603043034c43b5523b98f3e262a2e/packages/react-native-nitro-test/nitrogen/generated/ios/swift/HybridTestObjectSwiftKotlinSpec_cxx.swift#L863-L869

`__promise` is a reference-type (a Swift `class`). It itself is **not** `Sendable`.

According to [the Apple Developer docs on Sendable Functions and Closures](https://developer.apple.com/documentation/swift/sendable#Sendable-Functions-and-Closures), a `@Sendable` function cannot capture/use non-sendable reference types like `__promise`:

image

..but my Swift 5 compiler doesn't give a warning for this yet. So I don't know.

The problem is, I can't make the Swift `Promise` class `Sendable` (or an `actor`) because then I would have to start synchronizing everything around it's state (which _might_ be a good thing anyways, but adds overhead), and then we just dig deeper in the rabbit hole - all other Swift class types will have to be `Sendable` (or `actor`) too - which means every other `XYZHybridObjectSpec`, and `ArrayBufferHolder`, and `AnyMap`, and so on.

**Comment by mrousavy** [comment]
Honestly to me this Sendable bullshit is stupid.

It's so much DX overhead and code complexity to avoid a problem I rarely ever see - how often did you have data races when building iOS apps? 
Especially in SwiftUI things are so much easier around Threading.
If you really dispatch something to a new Thread (and aren't a bad developer) you think about data access anyways, and in most cases that can happen via queues, locks, mutexes or better yet atomics.

In our case it's even more stupid; we are coming from a single-threaded JS context (thru C++), so this is not really something that is concurrent anyways.

I feel like it shouldn't be Nitro's responsibility. I'll look into this more to understand it better.

**Comment by TimGass** [comment]
Yeah, as far as I can tell the problem is that Sendable type colors all closures and it seems to be very much so an all or nothing game. Swift 5 shouldn't give you those errors because their concurrency model doesn't force you to use them, but sadly, Swift 6 does. It could be a good thing in the end, but I imagine this could cause some serious backwards compatibility issues. I can't really see why this was the way Swift 6 wanted to roll out this feature, but maybe you have to either implement both options or just use `@preconcurrency` until a better option presents itself. Otherwise, it does seem like using Swift 6's concurrency model requires some serious redesigns of all data structures. I was hoping to not blow my request up into that big of a task, but it feels like there are no non-painful solutions here. I'd be happy with anything that just compiles in swift 6, but sorry if I kind of dropped this on you. I did not mean to create this much work, but it really feels like Swift 6 is breaking rank with a lot of the previous Swift versions. I'm perfectly happy with a middle of the road solution, but technically, a redesign of these structures would probably be the be the best way to "do it the Swift 6 way," I guess.

**Comment by mrousavy** [comment]
Let me revisit this once the new Xcode/MacOS is released for everyone, then Swift 6 is also the standard. I guess we can make everything that's sendable `@Sendable`, and any shared thingies `actor`s.

It will likely have an impact on performance tho I think.

**Comment by mrousavy** [comment]
You can workaround this issue quite easily tho I think.

```swift
// 1. you receive this from your method args
let nonSendableClosure = { in print("hi") }

// 2. wrap it in a @unchecked Sendable
struct SendableWrapper: @unchecked Sendable {
  let closure: () -> Void
  func call() {
    closure()
  }
}
let wrapper = SendableWrapper(closure: nonSendableClosure)

// 3. now you have a @Sendable closure
let sendableClosure = { @Sendable () in wrapper.call() }
funcThatTakesASendable(sendableClosure)
```

TimGass:
> Honestly to me this Sendable bullshit is stupid.
> 
> It's so much DX overhead and code complexity to avoid a problem I rarely ever see - how often did you have data races when building iOS apps? Especially in SwiftUI things are so much easier around Threading. If you really dispatch something to a new Thread (and aren't a bad developer) you think about data access anyways, and in most cases that can happen via queues, locks, mutexes or better yet atomics.
> 
> In our case it's even more stupid; we are coming from a single-threaded JS context (thru C++), so this is not really something that is concurrent anyways.
> 
> I feel like it shouldn't be Nitro's responsibility. I'll look into this more to understand it better.

I do kind of agree. I think the compiler is forcing people into scenarios that do not make sense. I think using `@preconcurrency` to kill the compiler check on it seems like a perfectly valid option, given that JS callbacks are single-threaded in the first place, but Swift is likely to remove it in the future at the rate they are headed. The good news is that there seems to be backlash from everyone having to rewrite their libraries, so perhaps we can hope that Swift realizes that this is an over-reach and relaxes the compiler in the future? I don't think you are responsible for this. I was just trying to find a minimal solution for getting the code to compile. I am really sorry that this is happening to any of us. It really feels like Swift made this decision with little to no concern for interoping with other languages.

TimGass:
> You can workaround this issue quite easily tho I think.
> 
> ```swift
> // 1. you receive this from your method args
> let nonSendableClosure = { in print("hi") }
> 
> // 2. wrap it in a @unchecked Sendable
> struct SendableWrapper: @unchecked Sendable {
>   let closure: () -> Void
>   func call() {
>     closure()
>   }
> }
> let wrapper = SendableWrapper(closure: nonSendableClosure)
> 
> // 3. now you have a @Sendable closure
> let sendableClosure = { @Sendable () in wrapper.call() }
> funcThatTakesASendable(sendableClosure)
> ```

I'll give it a shot. I tried wrappers, but they did not like subsequent calls until I slapped them with the `@preconcurrency`.

mrousavy:
Wanna see some magic? 😉

```swift
@available(iOS 17.0.0, *)
struct SendableWrapper<R, each A>: @unchecked Sendable {
  private let closure: (repeat each A) -> R
  init(_ closure: @escaping (repeat each A) -> R) {
    self.closure = closure
  }
  
  @inline(__always)
  func callAsFunction(_ args: repeat each A) -> R {
    return self.closure(repeat each args)
  }
}

@available(iOS 17.0.0, *)
func doSomeMagic() {
  // 1. get your non-Sendable closures
  let nonSendableClosure1 = { () in print("HI!") }
  let nonSendableClosure2 = { (a: Int, b: Int) -> Int in a + b }
  
  // 2. wrap them using the generic magic `SendableWrapper` © Marc Rousavy
  let sendableClosure1 = SendableWrapper(nonSendableClosure1)
  let sendableClosure2 = SendableWrapper(nonSendableClosure2)
  
  // 3. those thingies are now Sendable! (@unchecked, but Sendable!)
  sendableClosure1()
  let result2 = sendableClosure2(5, 7)
}
```

(only works on iOS 17, because variadic parameter packs (`repeat each`) was introduced later in Swift runtime)

TimGass:
> Wanna see some magic? 😉
> 
> ```swift
> @available(iOS 17.0.0, *)
> struct SendableWrapper<R, each A>: @unchecked Sendable {
>   private let closure: (repeat each A) -> R
>   init(_ closure: @escaping (repeat each A) -> R) {
>     self.closure = closure
>   }
>   
>   @inline(__always)
>   func callAsFunction(_ args: repeat each A) -> R {
>     return self.closure(repeat each args)
>   }
> }
> 
> @available(iOS 17.0.0, *)
> func doSomeMagic() {
>   // 1. get your non-Sendable closures
>   let nonSendableClosure1 = { () in print("HI!") }
>   let nonSendableClosure2 = { (a: Int, b: Int) -> Int in a + b }
>   
>   // 2. wrap them using the generic magic `SendableWrapper` © Marc Rousavy
>   let sendableClosure1 = SendableWrapper(nonSendableClosure1)
>   let sendableClosure2 = SendableWrapper(nonSendableClosure2)
>   
>   // 3. those thingies are now Sendable! (@unchecked, but Sendable!)
>   sendableClosure1()
>   let result2 = sendableClosure2(5, 7)
> }
> ```
> 
> (only works on iOS 17, because variadic parameter packs (`repeat each`) was introduced later in Swift runtime)

This is very cool! It kind of makes me wonder, which I have for a while now, what the difference between `@preconcurrency` and `@unchecked Sendable` really is when you boil down to it? There's a lot of mystery around these Sendables for me, lol. Swift really is one of the hardest to read languages I've ever used.

TimGass:
I got it to work. It's rough, but the ios 17 requirement isn't even bad considering this is a swift 6 problem for ios 18. This feels like typescript gymnastics all over again! The beauty of modern languages, I suppose.

TimGass:
@mrousavy 
If you are good with everything, I can close this pull request and shut down my fork. It seems like you have a good handle on how you want to implement any changes, if at all.

mrousavy:
Yea - let's do this over a feature request instead for now.

TimGass:
> Yea - let's do this over a feature request instead for now.

Sounds good. It's a good thing to keep on the horizon, though. Adding Sendable support would be nice in the future. Thanks for everything! This package is awesome!

mrousavy:
see https://github.com/mrousavy/nitro/issues/719

PR #718, opened 2025-07-08
feat: Mark all callbacks as `@Sendable`

mrousavy:
In Swift 6, Swift enforces Swift concurrency even more. Which is stupid imo if you think about interop with other languages (like JS or C++, which have different concurrency modules).

So either we mark all Swift callbacks as `@Sendable`, or we mark all as `@preconcurrency`.

- `@preconcurrency` seems like it would mark an API as legacy, which I don't think makes sense.
- `@Sendable` sounds like it just marks the function as callable from any Thread, which is what it essentially is.

That's at least the way I understood those concepts - it's not very well explained in low-level terminology imo, but let's just try if this works.

### DIFF
--- a/packages/nitrogen/src/syntax/swift/SwiftCxxBridgedType.ts
+++ b/packages/nitrogen/src/syntax/swift/SwiftCxxBridgedType.ts
@@ -87,7 +87,7 @@ export class SwiftCxxBridgedType implements BridgedType<'swift', 'c++'> {
         // SomeStruct (Swift extension) <> SomeStruct (C++)
         return true
       case 'function':
-        // (@ecaping () -> Void) <> std::function<...>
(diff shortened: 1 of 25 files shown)


TimGass:
My understanding is that `@preconcurrency` would essentially tell the compiler to not enforce the strict concurrency rules that require Sendable in that particular closure's nested closures to make it compatible with previous versions of Swift, which is probably not an ideal thing to disable. I think that you are generally correct in what `@Sendable` does. I believe it is doing something to enforce thread concurrency safety. It may be doing something like an atomically referenced counted mutex behind the hood, but I believe it is the generally proposed method by Swift 6 for promoting something like memory locking in closures to prevent race conditions, though I'd love more explanation from those that know.

PR #731, opened 2025-07-14
chore(example): fix iOS building issues

hannojg:
We need to keep the .xcode.env.local file for builds to work correctly.

Otherwise on a fresh setup we get errors like this during building:

CleanShot_2025-07-14_at_12 51 152x


mrousavy:
ty

PR #750, opened 2025-07-16
fix: Add test for two callbacks that take structs

mrousavy:


mrousavy:
Created an issue in the swift compiler: https://github.com/swiftlang/swift/issues/83097

mrousavy:
lol what, it does build in the CI but not locally! wtf?

Issue #674, opened 2025-05-24
Android: Cache issue when rebuilding, need to clear cache to fix


maitrungduc1410:
### What's happening?

I have my react-native-video-trim lib which uses Nitro modules: https://github.com/maitrungduc1410/react-native-video-trim

It works fine until I install `react-native-image-picker`. Then I run `yarn run android` and it's throwing error related to Nitro modules.

You can check my reproducible repo attached with this issue. 

I notice one thing, if I downgrade react-native-image picker to `7.1.2`, then it works. This is full log, it's quite long:

[log.txt](https://github.com/user-attachments/files/20426731/log.txt)

Thanks for your time

### Reproduceable Code

```tsx
Reproducible repo: https://github.com/maitrungduc1410/test-video-trim
```

### Relevant log output

```shell
Log is long, I attach it as file above
```

### Device

Samsung Galaxy S20 5G, Android 13

### Nitro Modules Version

^0.25.2

### Can you reproduce this issue in the Nitro Example app here?

No, I cannot reproduce the issue in the Example app

### Additional information

- [ ] I am using Expo
- [x] I am using Nitrogen (nitro-codegen)
- [x] I have read and followed the [Troubleshooting Guide](https://nitro.margelo.com/docs/troubleshooting).
- [ ] I created a reproduction PR to reproduce this issue here in the nitro repo. (See [Contributing](https://nitro.margelo.com/docs/contributing) for more information)
- [x] I searched for [similar issues in this repository](https://github.com/mrousavy/nitro/issues) and found none.

mrousavy:
Essentially these are your C++ errors:

```
  ld.lld: error: undefined symbol: margelo::nitro::HybridObject::HybridObject(char const*)
  >>> referenced by JHybridVideoTrimSpec.hpp:31 (/Users/ductrungmai/Code/react-native/TestVideoTrim/node_modules/react-native-video-trim/nitrogen/generated/android/c++/JHybridVideoTrimSpec.hpp:31)
[...]

PR #771, opened 2025-07-25
fix: Fix undefined symbols when syncing

mrousavy:
A super long standing issue in native dependencies of native dependencies (e.g. VisionCamera -> Worklets, or NitroImage -> Nitro) was that the second native dependency's symbols can no longer be linked after some time. E.g. after gradle syncs, or another npm install.

Apparently the guys at SWM found a fix for this (so huge shoutout to @tomekzaw!), and this needs to be applied to all third party dependencies.

- fixes https://github.com/mrousavy/nitro/issues/674

### DIFF
--- a/packages/react-native-nitro-test/android/build.gradle
+++ b/packages/react-native-nitro-test/android/build.gradle
@@ -145,3 +145,13 @@ if (isNewArchitectureEnabled()) {
     codegenJavaPackageName = "com.margelo.nitro.test"
   }
 }
+
+// This fixes linking errors due to undefined symbols from libNitroModules.so.
+// During Gradle Sync, Android Gradle Plugin runs Prefab and treats worklets
+// like a header-only library. During build, config files are not regenerated
+// because the cache key does not change and AGP thinks that they are up-to-date.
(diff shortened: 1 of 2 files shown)


tomekzaw:
Hey, thanks for the shoutout as well as mentioning the source of the workaround – we really appreciate this! ❤️

While this approach indeed fixes the issue, it's far from ideal, as it basically runs the build process during Gradle sync and it takes plenty of time for a thing that should take at most seconds.

The same issue appeared in Reanimated (after we moved Worklets to a separate dependency). I've tried to apply the very same approach in https://github.com/software-mansion/react-native-reanimated/pull/7781 but noticed that with the change Gradle sync takes couple of minutes which is unacceptable.

@lukmccall pointed out that prefab is supposed to be used to link prebuilds rather than for building the dependencies from source. He also suggested that we should avoid using prefab in favor of "manual" linking. We've submitted another draft PR that does exactly that: https://github.com/software-mansion/react-native-reanimated/pull/7866 Now we just need to declare a dependency between build tasks (so that Reanimated is built after Worklets) and perhaps find a better way to obtain the path for `libworklets.so` in debug/release mode once it's properly stripped etc.

mrousavy:
nvm we fixed this with this: https://github.com/mrousavy/nitro/commit/92abaaaac701110afa22f886e5144d5729f8b3a8

PR #798, opened 2025-08-08
perf: Speed up Nitrogen by 94x by making `filterDuplicateFiles(…)` O(n) instead of O(n²) 🤯

mrousavy:
1. I was walking the full array recursively, so it wasn't O(n) to dedupe files but rather O(n^2)!! 
2. I was comparing the `file.content`, which is an expensive string comparison (character by character) each time. Also called O(n^2) lol
3. Instead of recursive loops, I use a hash map to efficiently store files via their full paths. Makes things O(n) and fast.

### DIFF
--- a/packages/nitrogen/src/utils.ts
+++ b/packages/nitrogen/src/utils.ts
@@ -64,7 +64,7 @@ export function toUnixPath(p: string): string {
 }
 
 const sep = path.sep
-function unsafeFastJoin(...segments: string[]): string {
+export function unsafeFastJoin(...segments: string[]): string {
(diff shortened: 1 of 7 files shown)


packages/nitrogen/src/utils.ts
```diff
@@ -74,21 +74,24 @@ function getFullPath(file: SourceFile): string {
     file.platform,
     file.language,
     ...file.subdirectory,
-    file.content
+    file.name
```
mrousavy, line 77: this was a super dumb mistake i probably wrote this super late at night

PR #827, opened 2025-08-19
fix: Add `SWIFT_ENABLE_EXPLICIT_MODULES` flag

mrousavy:


mrousavy:
hmm nvm this flag will be enabled by default in Xcode 26 anyways. Better to just keep our build files simple.

PR #856, opened 2025-09-04
fix(kotlin): use primitive arrays (DoubleArray) instead of Array<Double>

patrickkabwe:
Fixes #854

Maps TypeScript arrays to Kotlin primitive arrays:
- `number[]` → `DoubleArray` (was `Array<Double>`)
- `boolean[]` → `BooleanArray` (was `Array<Boolean>`)
- `bigint[]` → `LongArray` (was `Array<Long>`)

**Before:**
```kotlin
val performanceScores: Array<Double>
```

**After:**
```kotlin
val performanceScores: DoubleArray
```

PR #858, opened 2025-09-04
test: Test native Kotlin array generation

mrousavy:


mrousavy:
yea patrick is right https://github.com/mrousavy/nitro/pull/856

PR #862, opened 2025-09-06
refactor: Ensure Objective-C compatibility by inheriting from NSObject

cloud2303:
This pull request updates the Swift hybrid object code generation logic to ensure Objective-C compatibility and clarify inheritance behavior. The most significant changes are focused on always inheriting from `NSObject` and making the inheritance intent explicit in the generated code.

#### Objective-C compatibility and inheritance logic

* The generated Swift base class now always inherits from `NSObject`, ensuring that all hybrid objects are compatible with Objective-C. If the object inherits from another hybrid base, `NSObject` is replaced by that base class, which already inherits from `NSObject`. [[1]](diffhunk://#diff-62334de10ab8282bd9a72d1582bee85fab4f1e02e9e43ac6e6f70d056f3e1b13L24-R25) [[2]](diffhunk://#diff-62334de10ab8282bd9a72d1582bee85fab4f1e02e9e43ac6e6f70d056f3e1b13L34-R57)
* The generated initializer is always overridden (`public override init() { super.init() }`) to make the inheritance intent explicit, regardless of whether a base class is present.
* The logic for overriding `getCxxWrapper` is clarified: it is only overridden if the object actually inherits from another generated hybrid base, not just `NSObject`.


mrousavy:
Hey - thank you for the PR, but sorry - no, this is not going into Nitro.

This adds overhead, puts a dependency on the Objective-C runtime, and is unnecessary for most people.

You can always create a separate class that inherits from NSObject to use that.

PR #866, opened 2025-09-09
perf: Cache `global.Promise` for faster Promise creations

mrousavy:
## Benchmarks

10.000 calls to `getValueFromJSCallbackAndWait`:

```ts
let promises: Promise<number>[] = []
for (let i = 0; i < ITERATIONS; i++) {
  promises.push(
    HybridTestObjectSwiftKotlin.getValueFromJSCallbackAndWait(() =>
      Math.random()
    )
  )
}
await Promise.all(promises)
const end = performance.now()
const nitroTime = end - start
```

- Before (main): ~160ms
- After (main): ~160ms

turns out this does not increase performance at all.

### DIFF
--- a/packages/react-native-nitro-modules/cpp/jsi/JSIConverter+Promise.hpp
+++ b/packages/react-native-nitro-modules/cpp/jsi/JSIConverter+Promise.hpp
@@ -32,7 +32,7 @@ struct JSIConverter<std::shared_ptr<Promise<TResult>>> final {
     auto thenCallback = [&]() {
       if constexpr (std::is_void_v<TResult>) {
         // void: resolve()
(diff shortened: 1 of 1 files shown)


mrousavy:
@ChristopherGabba I can reproduce the error here:

```diff
- ❌ Test "JS Promise<void> can be awaited on native side" failed! Error: Expected test to not throw any errors, but an error was thrown! Error: Error: Timeouted: Promise<void> was destroyed!
```

mrousavy:
nvm, it now works fine again. I broke it before, and now reverted my mistake in https://github.com/mrousavy/nitro/pull/866/commits/afab186105a3937cd548989d76818aacd66205d2.

PR #867, opened 2025-09-09
perf: Allow fast path for primitively copyable arrays

mrousavy:
Instead of iterating over each element, we can have a fast-path for Arrays that are primitively copyable - essentially this reduces the cost from `O(n + allocationCost)` to `O(allocationCost)`.
This affects the following arrays:
- `number[]`
- `boolean[]`
- `bigint[]`
- `enum[]`
- `struct[]`

### Swift -> C++

Before:

```swift
let __result = {
  var __vector = bridge.create_std__vector_Powertrain_(__array.count)
  for __item in __array {
    __vector.push_back(__item)
  }
  return __vector
}()
```

After:

```swift
let __result = bridge.copy_std__vector_Powertrain_(__array, __array.count)
```

### C++ -> Swift

Before:

```swift
let __result = array.map({ __item in __item })
```

After:

```swift
let __result = {
  let __data = bridge.get_data_std__vector_Powertrain_(array)
  let __size = array.size()
  return Array(UnsafeBufferPointer(start: __data, count: __size))
}()
```

### The Magic

It's mainly faster because we avoid iterating over each item one by one and copy-constructing it into the new array, but instead performing a trivial `memcpy` when possible:

```cpp
template <typename T>
std::vector<T> FastVectorCopy(const T* data, size_t size) {
  if constexpr (std::is_trivially_copyable_v<T>) {
    // FAST: Type does not have a copy constructor - simply memcpy it
    std::vector<T> vector;
    vector.reserve(size);
    std::memcpy(vector.data(), data, size * sizeof(T));
    return vector;
  } else {
    // SLOW: Type needs to be iterated to copy-construct it
    std::span<const T> span(data, size);
    return std::vector<T>(span.begin(), span.end());
  }
}
```

### Benchmarks

Code:

```ts
const ITERATIONS = 50_000

// Bounce a 5120 Double[] (5KB) back and forth between JS & Swift, 50.000 times
let array = Array<number>(1024 * 5)
  .fill(0)
  .map((_, i) => i)
for (let i = 0; i < ITERATIONS; i++) {
  array = HybridTestObjectSwiftKotlin.bounceNumbers(array)
}
const end = performance.now()
const nitroTime = end - start
```

Before: **12.986,48ms**

After: **15,14ms**

Improvement: 572x

### DIFF
--- a/packages/react-native-nitro-modules/cpp/utils/FastVectorCopy.hpp
+++ b/packages/react-native-nitro-modules/cpp/utils/FastVectorCopy.hpp
@@ -28,8 +28,7 @@ std::vector<T> FastVectorCopy(const T* data, size_t size) {
 
   if constexpr (std::is_trivially_copyable_v<T>) {
     // FAST: Type does not have a copy constructor - simply memcpy it
-    std::vector<T> vector;
-    vector.reserve(size);
+    std::vector<T> vector(size);
     std::memcpy(vector.data(), data, size * sizeof(T));
     return vector;
   } else {


mrousavy:
Okay I benchmarked more: 
Screenshot 2025-09-22 at 15 59 45


I ran the test on 5KB of data (1024 * 5) just indexed number arrays.
Once 5k times, once 50k times. On my mac.

### With the simple `for` loop, just like on current `main` branch

- 5k times: **1.356,46ms**
- 50k times: **13.354,57ms**

### With the `std::span` constructor for `std::vector`

- 5k times: **1.350,55ms**
- 50k times: **12.986,48ms**

### With `std::is_trivially_copyable_v` and a custom bulk `memcpy`

- 5k times: **2,36ms**
- 50k times: **15,14ms**

So apparently the memcpy thingy makes it **~572x faster** 👀

packages/react-native-nitro-test/nitrogen/generated/ios/swift/Car.swift
```diff
@@ -104,14 +100,7 @@ public extension Car {
...
     get {
-      return { () -> Person? in
-        if bridge.has_value_std__optional_Person_(self.__driver) {
-          let __unwrapped = bridge.get_std__optional_Person_(self.__driver)
-          return __unwrapped
-        } else {
-          return nil
-        }
-      }()
+      return self.__driver.value
```
mrousavy, line 103: this was an unnoticed bug before - structs can simply be unwrapped. I hope this doesn't break in any other cases, but it seems fine.

PR #871, opened 2025-09-14
fix: remove duplicate includes in hybrid view component generation

patrickkabwe:
This PR fixes duplicate include statements that were being generated in the C++ hybrid view component headers, improving code cleanliness and preventing potential compilation issues.

**Before/After**
Before: Generated headers contained duplicate includes like:

```cpp
#include <optional>
#include <NitroModules/NitroDefines.hpp>
// ... other includes
#include <optional>  // duplicate!
```

After: Clean headers with no duplicate includes:
```cpp
#include <optional>
#include <NitroModules/NitroDefines.hpp>
// ... other includes (no duplicates)
```

### DIFF
--- a/packages/nitrogen/src/views/CppHybridViewComponent.ts
+++ b/packages/nitrogen/src/views/CppHybridViewComponent.ts
@@ -75,8 +75,26 @@ export function createViewComponentShadowNodeFiles(
     (p) => `CachedProp<${p.type.getCode('c++')}> ${escapeCppName(p.name)};`
   )
   const cases = props.map((p) => `case hashString("${p.name}"): return true;`)
-  const includes = props.flatMap((p) =>
-    p.getRequiredImports('c++').map((i) => includeHeader(i, true))
+  // Collect and dedupe dynamic includes, excluding ones already added statically above
+  const dynamicIncludes = Array.from(
+    new Set(
+      props.flatMap((p) =>
+        p.getRequiredImports('c++').map((i) => includeHeader(i, true))
+      )
+    )
+  )
+  const fixedHeaderIncludes = new Set<string>([
+    '#include <optional>',
+    '#include <NitroModules/NitroDefines.hpp>',
+    '#include <NitroModules/NitroHash.hpp>',
+    '#include <NitroModules/CachedProp.hpp>',
+    '#include <react/renderer/core/ConcreteComponentDescriptor.h>',
+    '#include <react/renderer/core/PropsParserContext.h>',
+    '#include <react/renderer/components/view/ConcreteViewShadowNode.h>',
+    '#include <react/renderer/components/view/ViewProps.h>',
+  ])
+  const includes = dynamicIncludes.filter(
+    (inc) => !fixedHeaderIncludes.has(inc)
   )
 
   // .hpp code
@@ -86,14 +104,7 @@ ${createFileMetadataString(`${component}.hpp`)}
 
 #pragma once
 
-#include <optional>
-#include <NitroModules/NitroDefines.hpp>
-#include <NitroModules/NitroHash.hpp>
-#include <NitroModules/CachedProp.hpp>
-#include <react/renderer/core/ConcreteComponentDescriptor.h>
-#include <react/renderer/core/PropsParserContext.h>
-#include <react/renderer/components/view/ConcreteViewShadowNode.h>
-#include <react/renderer/components/view/ViewProps.h>
+${Array.from(fixedHeaderIncludes).join('\n')}
 
 ${includes.join('\n')}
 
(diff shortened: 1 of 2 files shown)


packages/nitrogen/src/views/CppHybridViewComponent.ts
```diff
@@ -75,9 +75,27 @@ export function createViewComponentShadowNodeFiles(
...
+  const fixedHeaderIncludes = new Set<string>([
+    '#include <optional>',
+    '#include <NitroModules/NitroDefines.hpp>',
+    '#include <NitroModules/NitroHash.hpp>',
+    '#include <NitroModules/CachedProp.hpp>',
+    '#include <react/renderer/core/ConcreteComponentDescriptor.h>',
+    '#include <react/renderer/core/PropsParserContext.h>',
+    '#include <react/renderer/components/view/ConcreteViewShadowNode.h>',
+    '#include <react/renderer/components/view/ViewProps.h>',
+  ])
```
mrousavy, line 88: This is very repetitive (`#include <...>` when it could just be the header name without the `#include `)

packages/nitrogen/src/views/CppHybridViewComponent.ts
```diff
@@ -75,9 +75,27 @@ export function createViewComponentShadowNodeFiles(
...
+  const dynamicIncludesSet = new Set<string>()
+  for (const prop of props) {
+    for (const import_ of prop.getRequiredImports('c++')) {
+      const include = includeHeader(import_, true)
+      if (!fixedHeaderIncludes.has(include)) {
+        dynamicIncludesSet.add(include)
+      }
+    }
+  }
+  const includes = Array.from(dynamicIncludesSet)
```
mrousavy, line 98: This overcomplicates the code a lot. For what, to avoid including `<optional>` twice? imo not worth it

packages/nitrogen/src/views/CppHybridViewComponent.ts
```diff
@@ -75,9 +75,27 @@ export function createViewComponentShadowNodeFiles(
...
+    '#include <NitroModules/NitroHash.hpp>',
+    '#include <NitroModules/CachedProp.hpp>',
+    '#include <react/renderer/core/ConcreteComponentDescriptor.h>',
+    '#include <react/renderer/core/PropsParserContext.h>',
+    '#include <react/renderer/components/view/ConcreteViewShadowNode.h>',
+    '#include <react/renderer/components/view/ViewProps.h>',
+  ])
+  const dynamicIncludesSet = new Set<string>()
+  for (const prop of props) {
+    for (const import_ of prop.getRequiredImports('c++')) {
```
mrousavy, line 91: `import_` -> name with a `_` suffix is not clean imo. I'd name it something different, like `requiredImport`.

packages/nitrogen/src/views/CppHybridViewComponent.ts
```diff
@@ -86,14 +104,7 @@ ${createFileMetadataString(`${component}.hpp`)}
...
 
-#include <optional>
-#include <NitroModules/NitroDefines.hpp>
-#include <NitroModules/NitroHash.hpp>
-#include <NitroModules/CachedProp.hpp>
-#include <react/renderer/core/ConcreteComponentDescriptor.h>
-#include <react/renderer/core/PropsParserContext.h>
-#include <react/renderer/components/view/ConcreteViewShadowNode.h>
-#include <react/renderer/components/view/ViewProps.h>
+${Array.from(fixedHeaderIncludes).join('\n')}
```
mrousavy, line 107: This is again an `Array.from` - I think we copy arrays/sets 4 times here. Not necessary a problem with performance, but code is not clean.

mrousavy (commented):
Thanks for your PR Patrick, but I don't think I will merge this because it overcomplicates the code a bit just to avoid including `<optional>` twice. This is not a problem in C++.

patrickkabwe:
> Thanks for your PR Patrick, but I don't think I will merge this because it overcomplicates the code a bit just to avoid including `<optional>` twice. This is not a problem in C++.

hey mrousavy  thanks for taking a look. I just wanted to clarify that this code isn’t only about preventing `<optional>` from being included twice, it also handles other types like `<functional>` and `<string>`. I’ve attached a screenshot for reference. let me know if this PR is still not needed

image

mrousavy:
Interesting, quite a lot of includes. I fixed this specific issue here: https://github.com/mrousavy/nitro/pull/874

mrousavy:
I guess the list can be added yea https://github.com/mrousavy/nitro/pull/874/commits/034bb2d07f53a3fc52ea2622324d4de7f6be4fed

This is a bit simpler imo tho. Lmk what you think

patrickkabwe:
> I guess the list can be added yea [034bb2d](https://github.com/mrousavy/nitro/commit/034bb2d07f53a3fc52ea2622324d4de7f6be4fed)
> 
> This is a bit simpler imo tho. Lmk what you think

I wasn’t aware there was a utility for excluding duplicates, your new PR definitely simplifies this implementation. I’ll close this one in favor of yours.

PR #872, opened 2025-09-14
fix: improve callback function type signature with conditional types

patrickkabwe:
This PR improves the `callback` helper function type signature by replacing function `overloads` with conditional types for better type inference.

**Issues with Original Overloads**
The original function overloads had type inference problems:

```typescript
// Before: Ambiguous overloads
export function callback<T extends (...args: any[]) => any>(func: T): { f: T }
export function callback<T>(func: T): T
```

**Problems:**
- TypeScript couldn't always determine which overload to use
- Inconsistent type inference for function parameters

**Screenshot**

image <br/>

**Solution**
Replaced with a single conditional type signature:

```typescript
// After: Single conditional type
export function callback<T>(
  func: T
): T extends (...args: any[]) => any ? { f: T } : T
```

**Test Cases**
```typescript
callback((x: number) => x)  // Returns { f: (x: number) => number }
callback(() => 1)           // Returns () => number
callback("hello")           // Returns "hello"
callback(42)                // Returns 42
```


mrousavy:
dope thanks!

Issue #883, opened 2025-09-17
Remove redundant Nitro logs for better UX


jpudysz:
### What's happening?

I think it would be great to remove these logs, since developers won’t know what they’re about. They could be behind an extra flag for debugging purposes or something similar. I was also approached by Kadi from Expo, who raised the same concern.

<img width="1000" alt="Image" src="https://github.com/user-attachments/assets/2be01a99-5ea3-4bda-ada6-7bc550e7e146" />

### Reproduceable Code

```tsx
Register any Hybrid on native side and check your JS logs.
```

### Relevant log output

```shell
[expoexample.debug.dylib] [INFO] [Nitro.JSICache] Destroying JSICache...
[expoexample.debug.dylib] [INFO] [Nitro.HybridObjectRegistry] Unregistering HybridObject "UnistylesRuntime"...
[expoexample.debug.dylib] [INFO] [Nitro.HybridObjectRegistry] Unregistering HybridObject "UnistylesStyleSheet"...
[expoexample.debug.dylib] [INFO] [Nitro.HybridObjectRegistry] Unregistering HybridObject "UnistylesShadowRegistry"...
[expoexample.debug.dylib] [INFO] [Nitro.HybridObjectRegistry] Registering HybridObject "UnistylesRuntime"...
[expoexample.debug.dylib] [INFO] [Nitro.HybridObjectRegistry] Successfully registered HybridObject "UnistylesRuntime"!
[expoexample.debug.dylib] [INFO] [Nitro.HybridObjectRegistry] Registering HybridObject "UnistylesStyleSheet"...
[expoexample.debug.dylib] [INFO] [Nitro.HybridObjectRegistry] Successfully registered HybridObject "UnistylesStyleSheet"!
[expoexample.debug.dylib] [INFO] [Nitro.HybridObjectRegistry] Registering HybridObject "UnistylesShadowRegistry"...
[expoexample.debug.dylib] [INFO] [Nitro.HybridObjectRegistry] Successfully registered HybridObject "UnistylesShadowRegistry"!
[expoexample.debug.dylib] [INFO] [Nitro.Dispatcher] Installing global Dispatcher Holder into Runtime "HermesRuntime[RNBridgeless] (com.facebook.react.runtime.JavaScript)"...
[expoexample.debug.dylib] [INFO] [Nitro.HybridObjectPrototype] Creating new JS prototype for C++ instance type "HybridNitroModulesProxy"...
[...]

PR #885, opened 2025-09-18
feat: Add `NITRO_MIN_LOG_LEVEL` compiler flag

mrousavy:
Adds the compiler flag `NITRO_MIN_LOG_LEVEL`.

This flag can be used to set the minimum log level for the `NitroLogger` - if it is set to `0` (`Debug`), all logs will be visible. If it is set to `2` (`Warning`) (the default), only warnings and errors will be visible, and debug-/info-logs will be dropped.

This flag can and should be set in your App's Podfile/CMake file. If not set, `2` will be the default. In the Nitro Example app we set it to `0`.

- Fixes https://github.com/mrousavy/nitro/issues/883

### DIFF
--- a/packages/react-native-nitro-modules/cpp/platform/NitroLogger.hpp
+++ b/packages/react-native-nitro-modules/cpp/platform/NitroLogger.hpp
@@ -16,6 +16,16 @@ namespace margelo::nitro {
 
 enum class LogLevel { Debug, Info, Warning, Error };
 
+constexpr LogLevel kMinLogLevel = [] {
+#ifdef NITRO_MIN_LOG_LEVEL
+    constexpr int v = NITRO_MIN_LOG_LEVEL; // passed via -DNITRO_MIN_LOG_LEVEL=0..3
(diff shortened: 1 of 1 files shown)


mrousavy:
Ah I forgot, since we are checking this in a Pod (and especially in a header), multiple pods all include this and run this check. So we need a way to add this flag to each pod, which is kinda ugly. Like this:

```rb
post_install do |installer|
  installer.pods_project.targets.each do |t|
    next unless t.name == 'YourPodName' # the pod target, not your app target
    t.build_configurations.each do |config|
      config.build_settings['GCC_PREPROCESSOR_DEFINITIONS'] ||= ['$(inherited)']
      config.build_settings['GCC_PREPROCESSOR_DEFINITIONS'] << 'LOG_LEVEL=2' # 0..3
      # or for C++ specifically:
      config.build_settings['OTHER_CPLUSPLUSFLAGS'] ||= ['$(inherited)']
      config.build_settings['OTHER_CPLUSPLUSFLAGS'] << '-DLOG_LEVEL=2'
    end
  end
end
```

I don't like that at all. I could move it into `nativeLog`, then it's only defined & checked once in the Nitro Modules core library

PR #891, opened 2025-09-19
fix: remove double spaces in nitrogen logs

AgustinMJ:
Removes unnecessary spaces when creating a Nitro Module using nitrogen (see screenshot below).

Screenshot 2025-09-19 at 15 21 36

If these spaces were intentional, feel free to close this PR. I just started trying Nitro Modules and noticed it. I'd love to learn and contribute to this project. Looks like a great, easy first contribution!

### DIFF
--- a/packages/nitrogen/src/init.ts
+++ b/packages/nitrogen/src/init.ts
@@ -13,7 +13,7 @@ export async function initNewNitroModule(
   ref: string = 'main'
 ): Promise<void> {
   Logger.info(
-    `⚙️  Creating new Nitro Module "${chalk.bold(moduleName)}" in ${chalk.underline(prettifyDirectory(baseDirectory))}...`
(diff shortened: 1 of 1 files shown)


mrousavy:
hm, some terminals render this incorrectly as 1-width instead of 2-width. I use macOS default Terminal.app, and without the double space some emojis render as one width, even if there is a space.

We could try if replacing the extra space with `\u200A` or `\u200B` (zero-width) works consistently for all terminals?

amillez:
> hm, some terminals render this incorrectly as 1-width instead of 2-width. I use macOS default Terminal.app, and without the double space some emojis render as one width, even if there is a space.
> 
> We could try if replacing the extra space with `\u200A` or `\u200B` (zero-width) works consistently for all terminals?

will have a look! thanks for checking!

PR #900, opened 2025-09-25
feat: Add `@final` JS Doc to make Hybrid Objects final

mrousavy:
By default, `HybridObject`s are subclassable. This PR adds an `@final` JS Doc to `HybridObjects` to make them final (therefore **not** subclassable):

```ts
interface A { value: number }
/** 
 * @final 
 */
interface B { value: number }
```

In this example, `A` is subclassable, while `B` is not.

### DIFF
--- a/docs/docs/hybrid-objects.md
+++ b/docs/docs/hybrid-objects.md
@@ -266,6 +266,18 @@ A Hybrid Object can either inherit from other Hybrid Objects, or satisfy a commo
   interface Video extends HybridObject, Media {}
   ```
 
+  ### Final HybridObjects
+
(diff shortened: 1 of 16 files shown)


mrousavy:
Did some benchmarks by running `addNumbers(5, 7)` together, once with and once without `final`:

<table>
<tr>
<th>Before (no <code>final</code>)</th>
<th>After (with <code>final</code>)</th>
</tr>
<tr>
<td><code>8.47ms</code></td>
<td><code>8.40ms</code></td>
</tr>
</table>

So a magnificent 0.8% improvement. Not even 1%. 

Closing because this just overcomplicates the codegen for no gain.

PR #905, opened 2025-09-26
docs: Generate custom `og:image` thumbnails per page

mrousavy:
Adds a docusaurus plugin that renders a custom SVG for each doc page using React, converts that to a PNG, then saves all of those as individual `og:image`s per doc page.

Now each page has a custom social thumbnail that contains it's title

<table>
<tr>
<th>Page</th>
<th>Before</th>
<th>After</th>
</tr>

<tr>
<td>Homepage</td>
<td>
og-card
</td>
<td>
og-card
</td>

<tr>
<td>How to build a Nitro Module</td>
<td>
og-card
</td>
<td>
og-card
</td>

<tr>
<td>Configuration</td>
<td>
og-card
</td>
<td>
og-card
</td>
</table>

### DIFF
--- a/docs/src/plugins/generate-og-images/NitroOgCard.tsx
+++ b/docs/src/plugins/generate-og-images/NitroOgCard.tsx
@@ -1,7 +1,7 @@
 import React from "react";
 
 const BORDER_WIDTH = 5
-const BORDER_RADIUS = 15
+const BORDER_RADIUS = 17
 const LAYERS_SPACING = 29
 
 const BLACK = 'rgb(34, 42, 65)'
--- a/docs/src/plugins/generate-og-images/plugin.tsx
+++ b/docs/src/plugins/generate-og-images/plugin.tsx
@@ -49,7 +49,7 @@ async function renderCard({ fonts, width, height, cardConfig, outputPath }: Rend
   await fs.mkdir(directory, { recursive: true })
   console.log(`Writing file "${outputPath}"...`)
   await fs.writeFile(outputPath, png)
-  console.log('Done!')
+  console.log(`Successfully generated card: ${outputPath}`)
 }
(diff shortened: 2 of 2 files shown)


slorber:
I'll review this soon, but this plugin might cause React hydration mismatches since the JS bundles are not in sync with what's in your HTML files (since you modify them with Cheerio after build => not a good idea)

We'll look into how to make it easier to achieve in Docusaurus, but here's how I would address the problem in userland today, which should already work with Docusaurus v3.9:

```ts
export default {
  markdown: {
    parseFrontMatter: async (params) => {
      // Reuse the default parser logic
      const result = await params.defaultParseFrontMatter(params);

      // Automatically generate a social card if a frontMatter.socialCardTitle is present in any MDX file
      if (result.frontMatter.socialCardTitle) {

        // Your responsibility is to write this function
        // You probably want to generate in website/static, add a cache, and disable this in dev
        const imagePath = await generateSocialCardWithCaching({
          title: result.frontMatter.socialCardTitle
        })

        // See https://docusaurus.io/docs/api/plugins/@docusaurus/plugin-content-docs#image
        result.frontMatter.image = imagePath
      }

      return result;
    },
  },
};
```

mrousavy:
Thanks for your reply Sebastien!

> [...] but this plugin might cause React hydration mismatches [...]

Hm I see. How does a React hydration mismatch show itself? Like, how does it affect UX? not sure if it currently has that issue with my approach, but I agree that it is a bit hacky.

> [...] here's how I would address the problem in userland today [...]

In your example, `socialCardTitle` is something I'd define at the top of each markdown file myself?
I guess I don't want to do that, I want to just use the `title` (and maybe `description`) of each page which gets automatically picked.

PR #906, opened 2025-09-28
chore: Make Android ViewManager class open for inheritance

patrickkabwe:
I'm currently working on [NitroText](https://github.com/patrickkabwe/react-native-nitro-text) support for Android and trying to override manager style methods like `setPadding` and more. Currently, since the generated class is `final`, I can't extend it to customize the behavior. With this change, I'll be able to go ahead and implement the custom styling I need.
This change modifies `class ${manager}` to `open class ${manager}` in generated Kotlin code, making all Nitro HybridView ViewManagers inheritable with no breaking changes - it's fully backward compatible.

**Before (not possible):**
```kotlin
// This would fail to compile
class NitroTextViewManager : HybridNitroTextManager() { // ❌ Error: This type is final
    override fun setPadding(view: View, left: Int, top: Int, right: Int, bottom: Int) {
        ViewCompat.setPaddingRelative(view, left, top, right, bottom)
    }
}
```

**After (now possible):**
```kotlin
// This compiles
class NitroTextViewManager : HybridNitroTextManager() { // ✅ Works!
    override fun setPadding(view: View, left: Int, top: Int, right: Int, bottom: Int) {
        ViewCompat.setPaddingRelative(view, left, top, right, bottom)
    }
}
```
This enables easy customization of NitroText styles and behavior by extending the generated ViewManager.

### DIFF
--- a/packages/react-native-nitro-test/nitrogen/generated/android/kotlin/com/margelo/nitro/test/views/HybridTestViewManager.kt
+++ b/packages/react-native-nitro-test/nitrogen/generated/android/kotlin/com/margelo/nitro/test/views/HybridTestViewManager.kt
@@ -17,7 +17,7 @@ import com.margelo.nitro.test.*
 /**
  * Represents the React Native `ViewManager` for the "TestView" Nitro HybridView.
  */
-class HybridTestViewManager: SimpleViewManager<View>() {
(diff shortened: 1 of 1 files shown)


mrousavy:
nice catch thanks patrick!

PR #910, opened 2025-09-29
Use absolute url so caching and prefetching works

tobiaslins:


mrousavy:
Es worked noch immer nicht 😭 


https://github.com/user-attachments/assets/46beeaf4-0bee-417e-a8e3-80f797e27e1d

PR #920, opened 2025-10-07
Reproduce: Error messages lost on iOS with react-native-skia

kalem-edlin:
- Add @shopify/react-native-skia to devDependencies
  - Add HelloWorld screen to demonstrate Skia is loaded
  - Run tests to see error messages are replaced with generic errors


mrousavy:
will be fixed in skia

PR #925, opened 2025-10-09
fix: optional variant fields dont use typescript name in generated files

### LINKED ISSUE #921: Optional variant/union fields don't use TypeScript type name in generated Swift enums
### What happened?

While working on react-native-health, I noticed that when an optional field is a variant (i.e., a filter field that is a large union of possible filters), the generated Swift enum name doesn't match the TypeScript type name.

I found a workaround: if you create another union that explicitly includes undefined with the optional field, it generates correctly.

In the example below nitrogen doesn't generate the name based on the type name (`Variant_DateFilter_DurationFilter` for the optional `DateOrDurationFilter` field).  For the `WorkaroundFilter` the generated name is correct and the file contents are the same as `Variant_DateFilter_DurationFilter`.

This has became a problem for me as I've hit the file name limits when the generated name was too long for Swift to compile.

### Nitrogen CLI logs (with `--logLevel="debug"`)

```tsx
🔧  Loading nitro.json config...
🚀  Nitrogen 0.0.1 runs at ~/werk/nitro-modules-reproduction/reproduction
    🔍  Nitrogen found 1 spec in ./src/specs
⏳  Parsing Example.nitro.ts...
    ⚙️   Generating specs for HybridObject "Example"...
        shared: Generating C++ code...
          C++: Creating HybridExampleSpec.hpp...
          C++: Creating HybridExampleSpec.cpp...
          C++: Creating OptionsWithRequired.hpp...
          C++: Creating DateFilter.hpp...
          C++: Creating DurationFilter.hpp...
          C++: Creating OptionWithWorkaround.hpp...
          C++: Creating OptionWithOptional.hpp...
        ios: Generating Swift code...
          Swift: Creating HybridExampleSpec.swift...
          Swift: Creating OptionsWithRequired.swift...
          Swift: Creating DateOrDurationFilter.swift...
          Swift: Creating DateFilter.swift...
          Swift: Creating DurationFilter.swift...
          Swift: Creating OptionWithWorkaround.swift...
          Swift: Creating WorkaroundFilter.swift...
          Swift: Creating OptionWithOptional.swift...
          Swift: Creating Variant_DateFilter_DurationFilter.swift...
          Swift: Creating HybridExampleSpec_cxx.swift...
          C++: Creating HybridExampleSpecSwift.hpp...
          C++: Creating HybridExampleSpecSwift.cpp...
        android: Generating Kotlin code...
          Kotlin: Creating HybridExampleSpec.kt...
          C++: Creating JHybridExampleSpec.hpp...
          C++: Creating JHybridExampleSpec.cpp...
          Kotlin: Creating OptionsWithRequired.kt...
          C++: Creating JOptionsWithRequired.hpp...
          Kotlin: Creating DateOrDurationFilter.kt...
          C++: Creating JDateOrDurationFilter.hpp...
          C++: Creating JDateOrDurationFilter.cpp...
          Kotlin: Creating DateFilter.kt...
          C++: Creating JDateFilter.hpp...
          Kotlin: Creating DurationFilter.kt...
          C++: Creating JDurationFilter.hpp...
          Kotlin: Creating OptionWithWorkaround.kt...
          C++: Creating JOptionWithWorkaround.hpp...
          Kotlin: Creating WorkaroundFilter.kt...
          C++: Creating JWorkaroundFilter.hpp...
          C++: Creating JWorkaroundFilter.cpp...
          Kotlin: Creating OptionWithOptional.kt...
          C++: Creating JOptionWithOptional.hpp...
          Kotlin: Creating Variant_DateFilter_DurationFilter.kt...
          C++: Creating JVariant_DateFilter_DurationFilter.hpp...
          C++: Creating JVariant_DateFilter_DurationFilter.cpp...
⛓️   Setting up build configs for autolinking...
    Creating autolinking build setup for ios...
          Ruby: Creating NitroReproduction+autolinking.rb...
          C++: Creating NitroReproduction-Swift-Cxx-Bridge.hpp...
          C++: Creating NitroReproduction-Swift-Cxx-Bridge.cpp...
          C++: Creating NitroReproduction-Swift-Cxx-Umbrella.hpp...
    Creating autolinking build setup for android...
          Cmake: Creating NitroReproduction+autolinking.cmake...
          Gradle: Creating NitroReproduction+autolinking.gradle...
          C++: Creating NitroReproductionOnLoad.hpp...
          C++: Creating NitroReproductionOnLoad.cpp...
          Kotlin: Creating NitroReproductionOnLoad.kt...
🎉  Generated 1/1 HybridObject in 0.8s!
💡  Your code is in ./nitrogen/generated
```

### Nitro Specs (`.nitro.ts`)

```ts
import type { HybridObject } from "react-native-nitro-modules";

interface DateFilter {
  startDate?: Date;
  endDate?: Date;
}

interface DurationFilter {
  duration: number;
}
type DateOrDurationFilter = DateFilter | DurationFilter

type WorkaroundFilter = DateOrDurationFilter | undefined

interface OptionsWithRequired {
  filters: DateOrDurationFilter
}

interface OptionWithOptional {
  filters?: DateOrDurationFilter 
}

interface OptionWithWorkaround {
  filters?: WorkaroundFilter 
}

export interface Example extends HybridObject<{ios: "swift", android: "kotlin"}> {
  process(options: OptionsWithRequired, workaround: OptionWithWorkaround, optionOptional?: OptionWithOptional, ): number
}
```

### Nitro Modules Version

0.29.8

### Nitrogen Version

0.29.8

### Operating system

MacOS

### Additional information

szymonograbek:
Fixes #921 

I've done some debugging on this issue and I've found out the following:

When an optional property or parameter uses a type alias for a variant, `getTypeAtLocation()` expands it to include `| undefined`, which creates a new anonymous union that loses the original alias name. This caused generated files to use names like `Variant_String_Car` instead of `NamedVariant`.

The fix uses `getTypeNode().getType()` to read the type from the syntax tree to preserve the alias name.

As a note, I'm not an expert on ts-morph or typescript's AST, so I don't know if this is the proper way to fix the original issue.

### DIFF
--- a/packages/react-native-nitro-test/cpp/HybridTestObjectCpp.hpp
+++ b/packages/react-native-nitro-test/cpp/HybridTestObjectCpp.hpp
@@ -147,7 +147,7 @@ public:
   void jsStyleObjectAsParameters(const JsStyleStruct& params) override;
   WrappedJsStruct bounceWrappedJsStyleStruct(const WrappedJsStruct& value) override;
   OptionalWrapper bounceOptionalWrapper(const OptionalWrapper& wrapper) override;
-  OptionalVariantWrapper bounceOptionalWrapperWithVariant(const OptionalVariantWrapper& wrapper) override; 
+  OptionalVariantWrapper bounceOptionalWrapperWithVariant(const OptionalVariantWrapper& wrapper) override;
   std::shared_ptr<ArrayBuffer> createArrayBufferFromNativeBuffer(bool /* copy */) override;
   std::shared_ptr<ArrayBuffer> createArrayBuffer() override;
   std::shared_ptr<ArrayBuffer> copyBuffer(const std::shared_ptr<ArrayBuffer>& buffer) override;


mrousavy:
Hey thanks for your PR, this is great!

I also really appreciate that you added a test case for this - two change requests I have before I can merge this:

1. Do you think we could simplify the logic in nitrogen to avoid having multiple different paths? Maybe we want to focus on only your path, and throw if the path cannot be hit (e.g. if `getTypeNode()` is undefined, which I guess should never be the case? Or maybe there's a different way that doesn't contain `undefined` that works in all situations...)
2. Could you add the two test functions to `getTests(...)` (`getTests.ts`) in the example too? Otherwise I can do that later.

szymonograbek:
@mrousavy done
I've only left the path I've added (`getTypeNode()`) - I've added some guards for safety and to not have typescript errors, although we shouldn't ever run into them. The only way I found to trigger the error is to use getter/setter in the interface:
```ts
interface Test {
  get foo(): string
  set foo(value: string)
}
```

szymonograbek:
Ah, I've found one issue tho, if in the `TestObject.nitro.ts` we change the property declaration to method signature in `JsStyleStruct` it throws the error:
```ts
interface JsStyleStruct {
  value: number
  onChanged(num: number): void
}
```

This again requires me to add a new path to the `getInterfaceProperties` to check if the property is a `MethodSignature` as `getTypeNode` doesn't exist on `MethodSignature`

EDIT: 
I've re-added the check for syntax kind

mrousavy:
`struct`s cannot have Method signature properties. That's intentional, and should throw an error.
Only function properties

szymonograbek:
> `struct`s cannot have Method signature properties. That's intentional, and should throw an error. Only function properties

Ok, I've restored the simplified approach

szymonograbek:
Looks like I've left a trailing white space so that made linting fail, I've removed it

Issue #922, opened 2025-10-08
Double conversion issue


pinpong:
Double conversion not working:

```js
type RNLatLng = { latitude: number; longitude: number };
```

with v0.29.6
```c++
  var coordinates: [RNLatLng] {
    @inline(__always)
    get {
      return self.__coordinates.map({ __item in __item })
    }
    @inline(__always)
    set {
      self.__coordinates = { () -> bridge.std__vector_RNLatLng_ in
        var __vector = bridge.create_std__vector_RNLatLng_(newValue.count)
        for __item in newValue {
          __vector.push_back(__item)
        }
        return __vector
      }()
    }
  }
````

result: 
latitude    Double    37.776826906253326
longitude    Double    -122.41762517753632

with v0.29.8
```c++
  var coordinates: [RNLatLng] {
    @inline(__always)
    get {
      return { () -> [RNLatLng] in
        let __data = bridge.get_data_std__vector_RNLatLng_(self.__coordinates)
        let __size = self.__coordinates.size()
        return Array(UnsafeBufferPointer(start: __data, count: __size))
      }()
    }
    @inline(__always)
    set {
      self.__coordinates = newValue.withUnsafeBufferPointer { __pointer -> bridge.std__vector_RNLatLng_ in
        return bridge.copy_std__vector_RNLatLng_(__pointer.baseAddress!, newValue.count)
      }
    }
  }
```
result:
latitude    Double    1.0358283582757498E-309
longitude    Double    2.569141358374482E-322

mrousavy:
Hey - I tried to reproduce your bug here in this PR; https://github.com/mrousavy/nitro/pull/927

But it doesn't fail. It works just fine. Can you try to reproduce this yourself somehow?

mrousavy:
I think https://github.com/mrousavy/nitro/pull/932 will fix this issue.

pinpong:
Na, it's not fixing the issue.
You can checkout [react-native-google-maps-plus](https://github.com/pinpong/react-native-google-maps-plus) and set breakpoint [here](https://github.com/pinpong/react-native-google-maps-plus/blob/main/ios/MapPolygonBuilder.swift#L8)
Before install update the nitrogen version to latest and run ```yarn nitrogen```
Within the example app navigate to Polygon test page and check the breakpoint

mrousavy:
I'm not cloning other example apps - we need to reproduce this here in the Nitro Example app.
Without a repro here, I cannot merge a fix

[...]

PR #927, opened 2025-10-09
reproduce `RNLatLng` vector issue

mrousavy:
This PR tries to reproduce https://github.com/mrousavy/nitro/issues/922 - but it just passes and works. No bug.

Screenshot 2025-10-09 at 13 11 00


mrousavy:
The author of the issue (@pinpong) reported that he receives weird values - I first thought that my "unsafe" array copy of the structs was not a good idea (endianness?), but apparently that works just fine. For comparison, I logged the entire array in JS and in Swift code to check both ends - and they all contain the same value:

JS:

```
[ { latitude: 37.776826906253326, longitude: -122.41762517753632 } ]
```

Swift:

```
[__C.margelo.margelo::nitro.margelo::nitro::test.margelo::nitro::test::RNLatLng(__latitude: 37.776826906253326, __longitude: -122.41762517753632)]
```

Issue #930, opened 2025-10-09
[swift] nested objects crashing or props missing


g4rb4g3:
### What's happening?

Hello again and sorry for bugging you again 🙂 

I'm facing an issue on swift side that the app either crashes or the newly added test for the example app does not work out.
When trying to access config.sections?.first?.items.first?.title.text on swift I don't get the proper value while on c++ it seems to be fine, at least when debugging it with xcode is see this:

<img width="949" height="916" alt="Image" src="https://github.com/user-attachments/assets/9d0c343f-54a3-46c6-8eba-3b9967cffea1" />

The test I've added expects `"1-section title-3-#1"` from the nestedObjectTest call but when running on iOS I get only `"1-section title-3-"` or a crash. On Android everything works out fine.

Crash:
<img width="1235" height="386" alt="Image" src="https://github.com/user-attachments/assets/33db6c66-0321-4fca-9d9a-7e48dd93cba0" />

### Reproduceable Code

```tsx
I did a quick and dirty test suite, removing the c++ stuff since I couldn't quickly figure out why nitrogen failed when I put it in the SharedTestObjectProps
But I guess c++ wouldn't be affected anyways since I see the proper values on the debugger.

https://github.com/g4rb4g3/nitro/tree/bug/swift-nested-objects
```

### Relevant log output

```shell
iOS/swift:
 (NOBRIDGE) LOG  ⏳ Test "HybridObject.nestedObjectTest is valid" started...
 (NOBRIDGE) LOG  1-section title-3-
 (NOBRIDGE) LOG  ❌ Test "HybridObject.nestedObjectTest is valid" failed! Error: Expected "false" (boolean) to equal "true" (boolean), but they are not equal!
 (NOBRIDGE) LOG  Showing Tests for HybridObject "TestObjectSwiftKotlin"


android/kotlin:
 (NOBRIDGE) LOG  ⏳ Test "HybridObject.nestedObjectTest is valid" started...
 (NOBRIDGE) LOG  1-section title-3-#1
 (NOBRIDGE) LOG  ✅ Test "HybridObject.nestedObjectTest is valid" passed!
 (NOBRIDGE) LOG  Showing Tests for HybridObject "TestObjectSwiftKotlin"
```

### Device

iOS 18.4 & 26.1 beta 2

### Nitro Modules Version

as provided by the example app

### Nitrogen Version

as provided by the example app

### Can you reproduce this issue in the Nitro Example app here?

[...]

PR #931, opened 2025-10-10
fix swift struct issue (#930)

g4rb4g3:
This PR fixes this issue https://github.com/mrousavy/nitro/issues/930
I checked the iOS example app and all tests pas just fine 😃


mrousavy:
Hey - thanks for your PR - this is the right fix, but I changed the logic a bit to refactor `isPrimitivelyCopyable` to make it more explicit, since `needsSpecialHandling` is technically a different thing. See https://github.com/mrousavy/nitro/pull/932

Thanks for your PR again! Huge help

PR #935, opened 2025-10-10
chore: Try to reproduce the struct issue with `NitroSection[]`

mrousavy:


mrousavy:
So I ran the test, and it works fine. It fails because the `onPress` functions are not equal, which is expected since they go through native to reach the Scheduler. But the array, struct and strings are converted just fine - no crash.

image

PR #932, opened 2025-10-10
fix: Make `structs` not primitively copyable

mrousavy:
- Fixes #930
- Fixes https://github.com/mrousavy/nitro/issues/922
- Fixes the breaking test that was added in https://github.com/mrousavy/nitro/pull/937

PR #936, opened 2025-10-10
add test for #932

g4rb4g3:
This PR adds a test for the swift struct issue we just fixed with #932.

Without the PR in place it fails (this is what this PR reflects):
image
With the fix it passes:
image
And it works flawless on Android no matter what 🙂 
image


mrousavy:
ok well this was a pain in the ass to reproduce. Managed to get a clean reproduction for this here: https://github.com/mrousavy/nitro/pull/937

Thanks for your help

g4rb4g3:
Well it was even more pain in the ass to face it while working on a library. 😱
Still we managed to fix it quite fast 😃

PR #967, opened 2025-10-17
feat: Create `ObjectUtils` for `Object.freeze`, `Object.defineProperty`, ...

mrousavy:
This one PR does 2 things:

## 1. Return actual error messages when Nitro fails to initialize

It logs & returns the actual error that was thrown when Nitro failed to initialize. This can happen if something is cooked with the `runtime`, or when a `global()` property cannot be set (e.g. if it's frozen)

So now the error would propagate up to JS instead of just saying the module cannot be found.

## 2. Add `ObjectUtils`

Also this PR adds `ObjectUtils`, a static-class that provides native access to JS methods like `Object.freeze`, `Object.create`, `Object.defineProperty` and a helper method `Object.defineGlobal`.

### Caching and better performance

`ObjectUtils` also uses caching internally for common JS functions (like `Object.defineProperty`), so this should speed things up as well.

### `Object.defineGlobal`

`Object.defineGlobal` takes a `KnownGlobalPropertyName` argument which is one of three values: `__nitroDispatcher`, `__nitroJsiCache` or `NitroModulesProxy` - all three values that Nitro puts into a `jsi::Runtime`'s `global`. As we add more to global, we also need to add them to the enum. This prevents typos as before it was just loose strings repeated everywhere.

`Object.defineGlobal` also adds DEBUG checks like freezing the global (prevent it from being written to) and throwing if it already exists. We had an issue where react-native-typegpu from @wcandillon also installed the same globals since he uses an old copy of Nitro, which causes duplicate code in the codebase. This causes undecypherable crashes later on, since we try to downcast our global types to our `nitro::margelo::Dispatcher` type (or other types), which obviously doesn't work since it is a different type (or more precisely: _C++ symbol_) - the one from webgpu.

### DIFF
--- a/packages/react-native-nitro-modules/src/turbomodule/NativeNitroModules.ts
+++ b/packages/react-native-nitro-modules/src/turbomodule/NativeNitroModules.ts
@@ -47,24 +47,6 @@ if (__DEV__) {
   }
 }
 
-// Freeze the globals we need
-if (__DEV__) {
-  function protectGlobal(key: keyof typeof globalThis): void {
-    const value = globalThis[key]
-    if (value == null) return
-    Object.freeze(value)
-    Object.defineProperty(globalThis, key, {
-      value: value,
-      writable: false,
-      configurable: false,
-    })
-  }
-
-  protectGlobal('__nitroDispatcher')
-  protectGlobal('__nitroJsiCache')
-  protectGlobal('NitroModulesProxy')
-}
-
 export function isRuntimeAlive() {
   const cache = globalThis.__nitroJsiCache
   const dispatcher = globalThis.__nitroDispatcher


wcandillon:
let me know if there is anything I can do on my side to quick fix the webgpu module (namespace renaming/global renaming)

mrousavy:
I think renaming the global would be enough for a monkey-patch, but nevertheless we should stop trying to monkeypatch all the time - going forward properly switching to the actually official release of Nitro makes more sense for both of us - from a maintenance and performance perspective :)

mrousavy:
Created a fix for WebGPU: https://github.com/wcandillon/react-native-webgpu/pull/271
let's merge this and then do nitro refactor

wcandillon:
@mrousavy nice! please trust that I not interesting to monkey patch forever, new bindings are high priority now.

PR #986, opened 2025-10-24
perf: Cache properties in a Swift struct

mrousavy:
Caches properties in a Swift struct.

Previously, accessing the same prop twice caused re-computation of the property (which essentially meant parsing from C++ to Swift again).
For large arrays, this meant a whole new deep copy of the entire array:
```swift
let first = myJSStruct.array
let second = myJSStruct.array
```

With this PR, the `second` will use the cached value from the first access.

Not sure if this is worth the complexity, just testing in a PR.


mrousavy:
ah nvm we cannot add properties to extensions.
should I make Swift structs a separate `struct` in Swift instead of re-using the C++ `struct` from Swift? I wonder if thats a good idea - like on Kotlin

Issue #977, opened 2025-10-21
`malloc` Crash when accessing a primitve array


g4rb4g3:
### What happened?

Hello again :)

I face a similar [issue](https://github.com/mrousavy/nitro/issues/930) we fixed with [0.30.0](https://github.com/mrousavy/nitro/releases/tag/v0.30.0)
When I try to access a Double array my app crashes.

<img width="1301" height="450" alt="Image" src="https://github.com/user-attachments/assets/41b2f07e-6a56-47cb-b893-7094dc149011" />

Commenting out [this if ](https://github.com/mrousavy/nitro/blob/main/packages/nitrogen/src/syntax/swift/SwiftCxxBridgedType.ts#L493)and always do what else does fixes the issue for me.

so instead of
```var angles: [Double] {
    @inline(__always)
    get {
      return { () -> [Double] in
        let __data = bridge.get_data_std__vector_double_(self.__angles)
        let __size = self.__angles.size()
        return Array(UnsafeBufferPointer(start: __data, count: __size))
      }()
    }
    @inline(__always)
    set {
      self.__angles = newValue.withUnsafeBufferPointer { __pointer -> bridge.std__vector_double_ in
        return bridge.copy_std__vector_double_(__pointer.baseAddress!, newValue.count)
      }
    }
  }
```

it generates
```var angles: [Double] {
    @inline(__always)
    get {
      return self.__angles.map({ __item in __item })
    }
    @inline(__always)
    set {
      self.__angles = newValue.withUnsafeBufferPointer { __pointer -> bridge.std__vector_double_ in
        return bridge.copy_std__vector_double_(__pointer.baseAddress!, newValue.count)
      }
    }
  }
```
making it not crash anymore.

### Nitrogen CLI logs (with `--logLevel="debug"`)

```tsx
nitrogen logs do not show anything related to this issue
```

### Nitro Specs (`.nitro.ts`)

```ts

```

### Nitro Modules Version

0.31.0

### Nitrogen Version

0.31.0

### Operating system

MacOS

### Additional information

- [x] I have read and followed the [Troubleshooting Guide](https://nitro.margelo.com/docs/troubleshooting).
- [ ] I created a reproduction PR to reproduce this issue here in the nitro repo. (See [Contributing](https://nitro.margelo.com/docs/contributing) for more information)
- [x] I searched for [similar issues in this repository](https://github.com/mrousavy/nitro/issues) and found none.

mrousavy:
[...]

PR #990, opened 2025-10-24
add test for #977

g4rb4g3:
This PR adds a failing/crashing swift test for https://github.com/mrousavy/nitro/issues/977 it works for c++ and kotlin though :)
Rarely it crashes, took me a few cold starts though to catch it.
Bildschirmfoto 2025-10-24 um 14 04 04

But even when it does not crash the test fails, printing each value of the array reveals that there is something terrible wrong, even though I pass in [100, 0] all the time I see different values on each attempt.
image

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -55,7 +55,7 @@
     },
     "example": {
       "name": "react-native-nitro-example",
-      "version": "0.31.0",
+      "version": "0.31.1",
       "dependencies": {
         "@react-native-segmented-control/segmented-control": "^2.5.7",
         "@react-navigation/bottom-tabs": "^7.4.5",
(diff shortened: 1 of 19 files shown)


g4rb4g3:
merging in latest main up until 7b542e0f3e80ade69d69281064a623db089dee55 makes it work.
image

mrousavy:
thanks! 👍

g4rb4g3:
merging in https://github.com/mrousavy/nitro/pull/984 also seems to fix it, did 10 cold starts with a few test runs each and it never crashed and always passed.
image

mrousavy:
fixed

PR #994, opened 2025-10-26
chore: Replace deprecated TurboReactPackage, ReactModuleInfo constructor 

l2hyunwoo:
스크린샷 2025-10-27 오전 1 35 53

스크린샷 2025-10-27 오전 1 36 47

remove warnings :fire:


mrousavy:
doesn't this break compatibility for older react-native versions?
Nitro promises to support RN 0.75+

mrousavy:
`BaseReactPackage` was introduced in https://github.com/facebook/react-native/pull/40775.

mrousavy:
Ok so this change is fine, but there a few other places where we need to replace this. e.g. docs.

PR #995, opened 2025-10-27
fix: Move `swiftPart` into ctor

mrousavy:


mrousavy:
This doesn't work because if a HybridObject extends another HybridObject, it also needs to pass the swiftPart up to the base.
We cannot move the same value twice, so we need to rely on copy.

Also, this is not a performance improvement since a Swift reference (`swift::_impl::RefCountedClass`) doesn't have move semantics - the C++ move constructor internally just does the same thing as the copy constructor (pass the opaque pointer and call `swift_retain`):

image

PR #1024, opened 2025-11-08
feat:- prevent-reserve-keyword to be used as property

riteshshukla04:


mrousavy:
There's properties, methods, HybridObject names, enum names, enum members and struct names - not sure if we want to check all of that...

riteshshukla04:
Agreed. It is not feasible to check everything. But Can we add it just for properties and methods (like my current PR), where majority of this cases lie.

Issue #1025, opened 2025-11-08
`npx nitrogen@latest init package` fails on win 11


psnet:
### What happened?

```
⏳ Cloning https://github.com/mrousavy/nitro...
fatal: Too many arguments.
...
Error: Command failed: git clone --depth 1 --filter=blob:none -b main --quiet --sparse https://github.com/mrousavy/nitro PATH_HERE
fatal: Too many arguments.
...
at downloadGitHubFolder
at initNewNitroModule
at async Object.handler
```

### Nitrogen CLI logs (with `--logLevel="debug"`)

```tsx
it failed at init, no logs
```

### Nitro Specs (`.nitro.ts`)

```ts
init failed, not specs
```

### Nitro Modules Version

idk

### Nitrogen Version

0.31.4

### Operating system

Windows

### Additional information

- [x] I have read and followed the [Troubleshooting Guide](https://nitro.margelo.com/docs/troubleshooting).
- [x] I created a reproduction PR to reproduce this issue here in the nitro repo. (See [Contributing](https://nitro.margelo.com/docs/contributing) for more information)
- [x] I searched for [similar issues in this repository](https://github.com/mrousavy/nitro/issues) and found none.

riteshshukla04:
Is there an extra space in your path? Maybe in your folder name or somewhere.
 This happens when there is a space in your absolute path .

psnet:
Yes, there is a space, lib should encode path?

riteshshukla04:
https://github.com/mrousavy/nitro/pull/1026 This should fix it

mrousavy:
amazing work @riteshshukla04 !

PR #1026, opened 2025-11-08
fix: nitrogen init with spaces in path

riteshshukla04:
Fixes https://github.com/mrousavy/nitro/issues/1025


mrousavy:
amazing thank you!

Issue #1050, opened 2025-11-19
HybridView is recycled and reused with old props for a new view on iOS


Phecda:
### What's happening?

I create a HybridView `MyCustomView` with props like
```
interface MyProps {
    propOptional?: string
}
```
1. use like `<MyCustomView propOptional="something" />`
2. unmount it
3. add another one like `<MyCustomView/>`

the second `MyCustomView` will use the same instance and receive the same propOptional `"something"` instead of undefined/null. I guess it is because on the Swift side the didSet was not triggered.

> I failed to run example app because `bundle install` failed

This can be reproduced by `resizeMode` of react-native-nitro-image / web-image (code see below). The current: B get unexpected resizeMode 'contain'

|  A | B |
| - | - |
|<img width="402" height="874" alt="Image" src="https://github.com/user-attachments/assets/064b02d1-2f4f-4786-96eb-982ca08d0c1e" /> | <img width="402" height="874" alt="Image" src="https://github.com/user-attachments/assets/43713df0-1d42-4036-920a-e33041d1265f" /> |

### Reproduceable Code

```tsx
function ImageA() {
  return (
    <NitroImage
      image={require('./9919.png')}
      style={{ backgroundColor: 'grey', width: 50, height: 100 }}
      resizeMode="contain"
      recyclingKey="A"
    />
  );
}

function ImageB() {
  return (
    <NitroImage
      image={require('./9919.png')}
      style={{ backgroundColor: 'grey', width: 100, height: 50 }}
      recyclingKey="B"
    />
  );
}

export default function () {
  const [isA, setIsA] = useState(true);
  return (
    <SafeAreaView style={styles.container}>
      <Switch value={isA} onValueChange={setIsA} />

      <Text>current: {isA ? 'A' : 'B'}</Text>
      {isA ? <ImageA /> : <ImageB />}

      <Divider />

      <Text>A:</Text>
      <ImageA />

      <Text>B:</Text>
      <ImageB />
    </SafeAreaView>
  );
}
```

### Relevant log output

```shell
No logs
```

### Device

iPhone 17 Pro simulator (iOS 26.1)

### Nitro Modules Version

0.31.4

### Nitrogen Version

0.31.4

### Can you reproduce this issue in the Nitro Example app here?

I didn't try (⚠️ your issue might get ignored & closed if you don't try this)

### Additional information

- [ ] I am using Expo
- [x] I am using nitrogen
[...]

PR #1055, opened 2025-11-21
[WIP] add ViewConfig arg to `NitroView`

hannojg:
Pass an option to nitro view to provide view config configuration.
Goal is to configure recycling support to help with issues as described here: 
- https://github.com/mrousavy/nitro/issues/1050


pinpong:
Nice, this can replace a part of my [patch](https://github.com/pinpong/react-native-google-maps-plus/blob/main/scripts/nitrogen-patch.js).

Phecda:
Any update?

PR #1057, opened 2025-11-22
feat: Create `Promise.parallel(qos: ...)`

mrousavy:


mrousavy:
not needed, you can always just also write `.global(qos: .default)` - not saving a lot of characters.

PR #1064, opened 2025-11-28
perf: Flatten structures of primitives into primitive Arrays in JNI

mrousavy:


mrousavy:
nvm this will not be faster because of array allocation.

what would be faster tho is calling the C++ converter from Kotlin, instead of the current other way around. that way we only have a single JNI call

PR #1065, opened 2025-11-28
fix: Fix optional doubles not building in Release mode on Swift

mrousavy:
Fixes a bug mentioned by https://github.com/hyochan/react-native-nitro-sound/pull/746#discussion_r2571562926 where a `number | undefined` (`std::optional<double>`) would not build in Release mode on Swift.

The fix is to not use the Swift generated `.value` accessor, but instead do a `.has_value() ? .pointer : nil` access.

The problem is, we use `std::optional<double>` in the `Person` struct in the example/test module, and it builds just fine in Release mode.

So this cannot be blindly merged until we have an actual test in this repo demonstrating this exact problem. cc @hyochan let me know if you find something

### DIFF
--- a/packages/nitrogen/src/syntax/swift/SwiftCxxBridgedType.ts
+++ b/packages/nitrogen/src/syntax/swift/SwiftCxxBridgedType.ts
@@ -471,7 +471,7 @@ export class SwiftCxxBridgedType implements BridgedType<'swift', 'c++'> {
 
             // TODO: Use `.value` instead of the `.has_value()`/`.pointee` once Swift has better support for `std::optional` bridging
             if (!wrapping.needsSpecialHandling && !swiftBug84848Workaround) {
-              return `${cppParameterName}.has_value() ? ${cppParameterName}.pointee : nil`
+              return `${cppParameterName}.hasValue ? ${cppParameterName}.pointee : nil`
             }
             return `
 { () -> ${optional.getCode('swift')} in
--- a/packages/react-native-nitro-test/nitrogen/generated/ios/swift/Car.swift
+++ b/packages/react-native-nitro-test/nitrogen/generated/ios/swift/Car.swift
@@ -124,7 +124,7 @@ public extension Car {
   var driver: Person? {
     @inline(__always)
     get {
-      return self.__driver.has_value() ? self.__driver.pointee : nil
+      return self.__driver.hasValue ? self.__driver.pointee : nil
     }
     @inline(__always)
     set {
--- a/packages/react-native-nitro-test/nitrogen/generated/ios/swift/Func_void_std__optional_double_.swift
+++ b/packages/react-native-nitro-test/nitrogen/generated/ios/swift/Func_void_std__optional_double_.swift
@@ -23,7 +23,7 @@ public final class Func_void_std__optional_double_ {
 
   @inline(__always)
   public func call(value: bridge.std__optional_double_) -> Void {
-    self.closure(value.has_value() ? value.pointee : nil)
+    self.closure(value.hasValue ? value.pointee : nil)
   }
 
   /**
--- a/packages/react-native-nitro-test/nitrogen/generated/ios/swift/HybridTestObjectSwiftKotlinSpec_cxx.swift
+++ b/packages/react-native-nitro-test/nitrogen/generated/ios/swift/HybridTestObjectSwiftKotlinSpec_cxx.swift
@@ -332,7 +332,7 @@ open class HybridTestObjectSwiftKotlinSpec_cxx {
     }
     @inline(__always)
     set {
-      self.__implementation.optionalEnum = newValue.has_value() ? newValue.pointee : nil
+      self.__implementation.optionalEnum = newValue.hasValue ? newValue.pointee : nil
     }
   }
   
@@ -989,7 +989,7 @@ open class HybridTestObjectSwiftKotlinSpec_cxx {
   @inline(__always)
   public final func tryOptionalEnum(value: bridge.std__optional_Powertrain_) -> bridge.Result_std__optional_Powertrain__ {
     do {
-      let __result = try self.__implementation.tryOptionalEnum(value: value.has_value() ? value.pointee : nil)
+      let __result = try self.__implementation.tryOptionalEnum(value: value.hasValue ? value.pointee : nil)
       let __resultCpp = { () -> bridge.std__optional_Powertrain_ in
         if let __unwrappedValue = __result {
           return bridge.create_std__optional_Powertrain_(__unwrappedValue)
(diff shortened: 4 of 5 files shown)


mrousavy:
This is where we have this case already: 
image


Why doesn't this fail in Release mode?

Maybe we need to actually access `.age` on the native Swift side for it to use the symbol? Not sure, but I can definitely use some help here to add a test to prevent future regressions.
I'm on vacation til monday

mrousavy:
This is where we have this case already: 
image


Why didn't this `.value` access fail in Release mode before? [All release builds in CI went through just fine](https://github.com/mrousavy/nitro/actions/workflows/build-ios-release.yml).

Maybe we need to actually access `.age` on the native Swift side for it to use the symbol? Not sure, but I can definitely use some help here to add a test to prevent future regressions.
I'm on vacation til monday

mrousavy:
I just added a test for this that passes the following struct:

```ts
interface PartialPerson {
  age?: number
  name?: string
}
```

..to native, and native Swift accesses `.age` and returns it to JS again.

I am running this test two times:
- with `age` being `undefined` (in which case `undefined` should be returned to JS)
- with `age` being `57` (in which case the number `57` should be returned to JS).

I built this in **release**, once with the code before:

```swift
  var age: Double? {
    @inline(__always)
    get {
      return self.__age.value
    }
```

..and once with the code after the change suggested by @hyochan in the thread over at his repo:

```swift
  var age: Double? {
    @inline(__always)
    get {
      return self.__age.hasValue ? self.__age.pointee : nil
    }
```

In **both** cases, the tests run fine without any corruption or errors in release mode: 
image

So we need another way to actually reproduce this. Are you guys maybe on an old Xcode version? I'm on **Version 26.1.1 (17B100)**.

hyochan:
I think there is still issue even we applied nitro patch in #746.

## Our Test Environment

- **Xcode: 26.1.1 (Build 17B100)** - Same as yours!
- React Native: 0.82.1
- react-native-nitro-modules: 0.31.10
- macOS: Darwin 24.6.0

## Our Test Results

We tested with the `.has_value() ? .pointee : nil` patch applied to our `AudioSet.swift`.

**AudioSet configuration in JS:**
```typescript
const audioSet = {
  AVSampleRateKeyIOS: 22050,
  AVNumberOfChannelsKeyIOS: 2,
  AudioSamplingRate: 22050,
  AudioChannels: 2,
  AudioEncodingBitRate: 64000,
};
```

**Release build recording result:**
```
Data format: 1 ch, 8000 Hz, Int16
```

Expected: 2 channels, 22050 Hz
Actual: 1 channel, 8000 Hz (fallback defaults)

The patch did NOT fix the issue for us in Release builds.

## Key Difference: Struct Complexity

Your `PartialPerson` test struct:
- 2 optional properties (`age?: number`, `name?: string`)
- Simple round-trip test: JS → Native → JS

Our `AudioSet` struct:
- **18 optional properties** (multiple `double?`, enum types, `boolean?`)
- Struct is passed to native and values are **consumed internally** (not returned to JS)
- Multiple optional properties accessed in sequence within native code

```json
{
  "AudioSet": {
    "AudioEncoderAndroid": "AudioEncoderAndroidType?",
    "OutputFormatAndroid": "OutputFormatAndroidType?",
    "AudioSourceAndroid": "AudioSourceAndroidType?",
    "AVEncoderAudioQualityKeyIOS": "AVEncoderAudioQualityIOSType?",
    "AVModeIOS": "AVModeIOSOption?",
    "AVEncodingOptionIOS": "AVEncodingOption?",
    "AVFormatIDKeyIOS": "AVEncodingOption?",
    "AVNumberOfChannelsKeyIOS": "double?",
    "AVSampleRateKeyIOS": "double?",
    "AVLinearPCMBitDepthKeyIOS": "AVLinearPCMBitDepthKeyIOSType?",
    "AVLinearPCMIsBigEndianKeyIOS": "boolean?",
    "AVLinearPCMIsFloatKeyIOS": "boolean?",
    "AVLinearPCMIsNonInterleavedIOS": "boolean?",
    "AudioQuality": "AudioQualityType?",
    "AudioChannels": "double?",
    "AudioSamplingRate": "double?",
    "AudioEncodingBitRate": "double?",
    "IncludeBase64": "boolean?"
  }
}
```

## Possible Causes

1. **Struct size/complexity**: The bug may only manifest with larger structs (18 vs 2 optional properties)
2. **Memory layout**: Swift compiler optimization may corrupt memory when many optionals exist in one struct
3. **Property access pattern**: Accessing multiple optional properties in sequence within native code (not round-trip)
4. **Mixed optional types**: Our struct has `double?`, `enum?`, and `boolean?` mixed together

## Suggestion for Reproduction

Could you try a more complex struct similar to ours?

```typescript
interface ComplexStruct {
  prop1?: number  // double?
  prop2?: number
  prop3?: number
  prop4?: number
  prop5?: SomeEnum  // enum type
  prop6?: SomeEnum
  prop7?: number
  prop8?: number
  prop9?: boolean
  prop10?: boolean
  prop11?: number
  prop12?: number
}
```

And access multiple properties in sequence on the native side (not just one property round-trip).

## How to Reproduce with Our Repo

```bash
git clone https://github.com/hyochan/react-native-nitro-sound.git
cd react-native-nitro-sound
yarn install
cd example
yarn ios:pod
yarn build:ios  # Release build
# Run on simulator, go to "NitroSound with Hook" screen, record audio
# Check recorded file with: afinfo <path_to_m4a_file>
```

The recorded file should show 22050 Hz, 2 channels if working correctly.
In Release build, it shows 8000 Hz, 1 channel (defaults).

mrousavy:
> Could you try a more complex struct similar to ours?

Can you try this? Fork this repo and just edit the structs - it's quite easy to build the example app.

I'm on vacation now


~~which Xcode version are you using?~~ EDIT: my bad, i didn't see - you are using the same version hmmm

mrousavy:
I think if you can reproduce this you should probably create a bug report in the swiftlang/swift repo.

hyochan:
> I think if you can reproduce this you should probably create a bug report in the swiftlang/swift repo.

I've tested current PR directly in our react-native-nitro-sound repo.
Here are the results:

### Test Setup:
- Installed nitrogen from PR #1065 branch via npm pack
- Ran yarn prepare to regenerate Swift files
- Built iOS Release for simulator

### Result:
afinfo output:
```
Data format: 1 ch, 8000 Hz, Int16
Expected: 2 ch, 22050 Hz
```

### Key Observation:

The generated [AudioSet.swift](https://github.com/hyochan/react-native-nitro-sound/blob/main/nitrogen/generated/ios/swift/AudioSet.swift) shows mixed patterns:
- Some properties use .hasValue (PR #1065 fix):
   `return self.__AVSampleRateKeyIOS.hasValue ? self.__AVSampleRateKeyIOS.pointee : nil`
- Some properties still use .has_value() (old pattern):
   `return self.__AVEncoderAudioQualityKeyIOS.has_value() ? self.__AVEncoderAudioQualityKeyIOS.pointee : nil`
- Bool properties use bridge helper functions:
   `if bridge.has_value_std__optional_bool_(self.__AVLinearPCMIsBigEndianKeyIOS) { ... }`

Our struct has 18 optional properties (vs your PartialPerson test with 2)

hyochan:
> I think if you can reproduce this you should probably create a bug report in the swiftlang/swift repo.

Done in https://github.com/swiftlang/swift/issues/85735. Have nice vacation!

PR #1066, opened 2025-11-30
feat(nitrogen): add error summary on generation failure

mfazekas:
Adds a clear summary at the end of nitrogen output when specs fail to generate, making failures harder to miss.

## Output comparison

|  | Success | Failure |
|--|---------|---------|
| **Before** | `...`<br>`🎉 Generated 13/13 HybridObjects...` | `...`<br>`⚙️ Generating specs for "RiveFile"...`<br>`    ❌ Failed to generate spec for RiveFile!`<br><br>`...` *(many more specs)* `...`<br><br>`🎉 Generated 12/13 HybridObjects...` |
| **After** | `...`<br>`✅ All specs generated successfully (13/13)`<br>`🎉 Generated 13/13 HybridObjects...` | `...`<br>`⚙️ Generating specs for "RiveFile"...`<br>`    ❌ Failed to generate spec for RiveFile!`<br><br>`...` *(many more specs)* `...`<br><br>`❌ Nitrogen generation failed`<br>`    • RiveFile: Error: The TypeScript type "any"...`<br>`    12 specs succeeded`<br>`🎉 Generated 12/13 HybridObjects...` |

The problem was that failures were easy to miss - the error appeared inline during generation and could scroll by unnoticed when processing many specs. Now there's always a clear summary at the end.

### DIFF
--- a/packages/nitrogen/src/nitrogen.ts
+++ b/packages/nitrogen/src/nitrogen.ts
@@ -18,7 +18,7 @@ import {
 import { writeFile } from './writeFile.js'
 import chalk from 'chalk'
 import { groupByPlatform, type SourceFile } from './syntax/SourceFile.js'
-import { Logger, type GenerationError } from './Logger.js'
+import { Logger } from './Logger.js'
 import { NitroConfig } from './config/NitroConfig.js'
 import { createIOSAutolinking } from './autolinking/createIOSAutolinking.js'
 import { createAndroidAutolinking } from './autolinking/createAndroidAutolinking.js'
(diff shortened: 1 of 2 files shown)


packages/nitrogen/src/nitrogen.ts
```diff
@@ -227,6 +232,13 @@ export async function runNitrogen({
     Logger.error(`❌  Failed to write ${chalk.dim(`.gitattributes`)}!`)
   }
 
+  // Display summary
```
mfazekas:
```suggestion
```

mrousavy:
on another note; do you think we should just abort/exit the CLI after it encountered an error?

mfazekas:
> on another note; do you think we should just abort/exit the CLI after it encountered an error?

Yes failing fast makes sense for nitrogen I think. The downside is that we'll only show the first error if there are multiple. i think I like failing fast a tiny bit more, but showing multiple errors is probably more efficient for LLM-s.  

So:  I don't know

PR #1072, opened 2025-12-01
test: add SelfReferentialStruct test case

puckey:
Adds a struct with a callback that references the same struct type.

Currently fails with "Maximum call stack size exceeded" due to infinite recursion in nitrogen's type walker.


mrousavy:
We need to add https://nitro.margelo.com/docs/types/custom-structs#cyclic-references-are-not-supported

puckey:
For this case I expected it to be possible, or is it something you don't want to support:

```ts
export interface SelfReferentialStruct {
  transform?: (config: SelfReferentialStruct) => Promise<SelfReferentialStruct>
}
```

puckey:
Closing in favor of #1074 which includes this test case along with a fix

PR #1073, opened 2025-12-01
fix: Reconstruct enum types from flattened string literal unions in variantsFix/boolean string enum variant

puckey:
When TypeScript flattens `boolean | SomeEnum` into `boolean | 'foo' | 'bar'`, the named enum type information is lost. This PR reconstructs the original enum types by searching all type aliases in the project and matching their string literal values.

## Changes
- Add `initializeTypeCreation()` to build a cache of type aliases for efficient lookup
- Add `reconstructEnumTypesFromStringLiterals()` to find matching enum types from flattened unions
- Handle variants with multiple string enums
- Detect and throw error for overlapping enums
- Added test cases in `TestObject.nitro.ts`
- Added documentation to variants.md

### DIFF
--- a/docs/docs/types/variants.md
+++ b/docs/docs/types/variants.md
@@ -16,9 +16,38 @@ While variants are still very efficient, they need runtime-checks for type conve
 which comes with a tiny overhead compared to all other statically defined types. If possible, **avoid variants**.
 :::
 
-## No literal values
+## Variants with String Enums
 
-A variant can only consist of types, not of literal values.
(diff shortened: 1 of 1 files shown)


mrousavy:
Hey - thank you so much for this PR, this is pretty cool.

I would say this is a very convenient feature, but I am a bit hesitant on merging it due to the amount of code it requires, and stuff like the "hacky" `initializeTypeCreation` global function.. I'll think about it :)

PR #1074, opened 2025-12-02
fix(nitrogen): handle cyclic struct references

puckey:
## Summary
  - Fixes stack overflow during code generation for structs with cyclic references
  - Throws a clear error for recursive struct types (struct containing itself directly or indirectly through other structs)
  - Adds support for structs with callbacks that reference the same struct type
  - Detects cyclic dependencies between struct and function types at code generation time
  - Generates separate `.hpp`/`.cpp` files for types with cyclic dependencies to avoid circular includes (Android)
  - Uses forward declarations in headers, with full type definitions only in implementation files
  - Adds `visited` set parameter to prevent infinite recursion in type walkers (affects all platforms)
  - Documents cyclic struct reference behavior in the docs

  ## Test plan
  - [x] Added `SelfReferentialStruct` test case with callback that returns the same struct type
  - [x] Added `TreeNode` and `TreeNodeMap` test cases for tree-like data structures
  - [x] Verified Android build succeeds
  - [x] Verified iOS build succeeds
  - [x] On-device tests pass for self-referential struct operations (Android + iOS)

  ---

  > [!NOTE]
  > This PR was implemented with Claude Code. I'm not deeply familiar with all the C++/JNI internals here, but did my best to follow along and verify the changes make sense. Also tested it in library code I am working on.


mrousavy:
Hey - I think overall the change is good - it's making DX better. I really appreciate you taking the time to work on this. :heart:

I am just not a big fan of how complex this is to implement - not only does it touch a lot of the codebase, but also it passes the `Set`s down very far where which many functions write to - it's a bit too complex for this small quality of life change imo.

We can leave this PR open for future reference, but I was hoping for a much smaller change...

puckey:
Thanks for the feedback! I've reworked this into two smaller, focused PRs:

- #1231 — fixes the exponential type re-traversal in `SwiftCxxBridgedType` / `KotlinCxxBridgedType` by adding visited-set guards to `getRequiredImports` and `getExtraFiles`. This turned out to be a separate pre-existing issue (#1079).
- #1232 — adds cyclic/self-referential struct support with a much simpler approach: lazy property resolution in `StructType` instead of passing `Set`s through the whole codebase.

Happy to close this one in favour of those.

PR #1075, opened 2025-12-02
fix(example): prevent "Nitro linked twice" error in monorepo

puckey:
## Summary
Use Metro's blockList to exclude duplicate copies of `react-native-nitro-modules` from nested `node_modules` in `packages/*/`. This ensures the module is only loaded once from the root, preventing the "NitroModules.install() was already called" error on Android.

Regression from 00513e5c which removed the exclusionList.

## Test plan
- [x] Verified Android example app runs without "Nitro linked twice" error
- [x] Verified iOS example app runs correctly


niteshbalusu11:
Thank you for this, took me forever to figure out what was going on.

PR #1076, opened 2025-12-02
fix(example): add dark mode text colors to test screens

puckey:
## Summary
Text was invisible on dark backgrounds in the example app. Add dynamic text colors using the existing `useColors()` hook to support both light and dark modes.

## Test plan
- [x] Verified text is visible in dark mode on iOS
- [x] Verified text is visible in dark mode on Android


puckey:
android_screenshot

mrousavy:
makes sense, thanks!

PR #1103, opened 2025-12-15
fix: Improve 16kb page size compatibility on android for NDK < 27

patrickkabwe:
I found that some Android environments using older NDK versions may encounter 16 KB page size issues due to incompatible default page sizes. Explicitly setting a 16 KB maximum page size improves compatibility across different NDK versions.
https://developer.android.com/guide/practices/page-sizes#compile-r26-lower

image

### DIFF
--- a/packages/react-native-nitro-modules/android/CMakeLists.txt
+++ b/packages/react-native-nitro-modules/android/CMakeLists.txt
@@ -47,6 +47,11 @@ find_library(LOG_LIB log)
 find_package(fbjni REQUIRED NitroConfig)
 find_package(ReactAndroid REQUIRED NitroConfig)
 
+# Set max page size to 16KB for older NDK versions
+if(CMAKE_ANDROID_NDK_VERSION VERSION_LESS "27")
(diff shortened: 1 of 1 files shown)


mrousavy:
NDK 26 or lower is deprecated / no longer supported, I don't think we want to support this

patrickkabwe:
> NDK 26 or lower is deprecated / no longer supported, I don't think we want to support this

Most react native apps are still using NDK 26 or earlier. Adding support for this will avoid the build issues that typically arise when updating to a newer NDK version(haven't tried this yet but have had a hard time in the past)

mrousavy:
react-native uses NDK 27 since react-native 0.77 - below that (0.76) it uses NDK 26. (https://github.com/react-native-community/template/blob/4c34b4bf9dc5825773a04ef0a46eb16c8e592337/template/android/build.gradle#L7)

I guess it's right around the point where we can drop support for it. I think it is easier to not push this onto maintainers, but onto app users instead - upgrade to RN 77, 78, 79, 80, 81, 82 or 83 instead. 
There's so many RN versions in-between.

patrickkabwe:
Makes sense. i will close this PR for now

Issue #1107, opened 2025-12-17
Android Segfault in ObjectUtils::create - Incorrect React Native version checks


jackprice:
### What happened?

On Android, the import path for `ReactNativeVersion.h` is incorrect. This means Android does not detect the React Native version correctly, and does not set `ENABLE_NATIVE_OBJECT_CREATE`.

There is some underlying bug that causes the code path taken in `margelo::nitro::ObjectUtils::create` with `ENABLE_NATIVE_OBJECT_CREATE` disabled to segfault on some devices, which we have not recreated, but can see from Sentry logs. I am unsure why this codepath causes segfaults, but it is only on Android that this codepath is taken, and only on Android we see these.

[I have a PR which resolves the import](https://github.com/mrousavy/nitro/pull/1108), which still builds and works, but we still haven't been able to reproduce the underlying issue so would appreciate any additional context anyone can add on these codepaths.

### Reproduceable Code

```tsx
On startup
```

### Relevant log output

```shell
All I have is stack traces from Sentry:

OS Version: Android 15 (CRT-N31 9.0.0.149(C185E5R2P3))
Report Version: 104

Exception Type: Unknown (SIGSEGV)

Application Specific Information:
Segfault

Thread 0 Crashed:
0   split_config.arm64_v8a.apk      0x7adae0698c        <unknown> + 527658150284
1   split_config.arm64_v8a.apk      0x7ad7512a90        margelo::nitro::ObjectUtils::create
2   split_config.arm64_v8a.apk      0x7ad74fd0cc        margelo::nitro::HybridObjectPrototype::createPrototype
3   split_config.arm64_v8a.apk      0x7ad74fd0b8        margelo::nitro::HybridObjectPrototype::createPrototype
4   split_config.arm64_v8a.apk      0x7ad74eb134        margelo::nitro::HybridObject::toObject
5   split_config.arm64_v8a.apk      0x7ad74fa814        margelo::nitro::install
6   split_config.arm64_v8a.apk      0x7ad74fa700        margelo::nitro::install
7   split_config.arm64_v8a.apk      0x7ad752a980        margelo::nitro::JNitroModules::install
8   split_config.arm64_v8a.apk      0x7ad752b644        facebook::jni::detail::MethodWrapper<T>::dispatch
9   split_config.arm64_v8a.apk      0x7ad752b574        facebook::jni::detail::FunctionWrapper<T>::call
10  boot.oat                        0x72045bf0          <unknown> + 1912888304
```

[...]

PR #1108, opened 2025-12-17
Use correct import path for ReactNativeVersion.h

jackprice:
When building for iOS, importing `React-cxxreact/cxxreact/ReactNativeVersion.h` works as expected. On Android, this path is not configured. It doesn't error because of the `#if __has_include`.

This means that on Android the React Native version check does not work, and so `ENABLE_NATIVE_OBJECT_CREATE` is not defined on Android.

See https://github.com/mrousavy/nitro/issues/1107


mrousavy:
Great stuff thanks!

PR #1109, opened 2025-12-18
perf: Add bulk `HashMap` conversion method to `AnyMap` to make conversion faster

riteshshukla04:
This makes toMap faster.

This is same as PR https://github.com/mrousavy/nitro/pull/1098. I have removed all benchmarks and made code cleaner

### DIFF
--- a/packages/react-native-nitro-modules/android/src/main/cpp/core/JAnyMap.hpp
+++ b/packages/react-native-nitro-modules/android/src/main/cpp/core/JAnyMap.hpp
@@ -15,46 +15,6 @@ namespace margelo::nitro {
 
 using namespace facebook;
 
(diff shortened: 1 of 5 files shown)


packages/react-native-nitro-modules/android/src/main/cpp/core/JAnyMap.hpp
```diff
@@ -15,6 +15,48 @@ namespace margelo::nitro {
 
 using namespace facebook;
 
+// Pre declre functions for recursive conversion
```
mrousavy:
```suggestion
// Forward declarations for recursive conversion
```

mrousavy:
thank you!!

PR #1115, opened 2025-12-18
perf: Avoid locks in `JSICache`

mrousavy:
Theoretically we don't need a lock in `JSICache`.

I'm still testing if this breaks under constant hot-reloads (or when runtimes tear down often, e.g. in react-native-worklets)


mrousavy:
doesn't make things faster

PR #1122, opened 2025-12-21
feat: add Harness

V3RON:
This pull request adds Harness, the E2E testing framework for React Native, focused on native module testing.

A placeholder test is included, which doesn’t add any value. Feel free to replace it with a proper test suite.

See https://www.react-native-harness.dev for details.

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -1,5 +1,6 @@
 {
   "lockfileVersion": 1,
+  "configVersion": 0,
   "workspaces": {
     "": {
       "name": "react-native-nitro",
@@ -55,7 +56,7 @@
     },
     "example": {
       "name": "react-native-nitro-example",
-      "version": "0.32.0-beta.0",
+      "version": "0.32.0",
       "dependencies": {
         "@react-native-segmented-control/segmented-control": "^2.5.7",
         "@react-navigation/bottom-tabs": "^7.4.5",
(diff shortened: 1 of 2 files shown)


V3RON:
@mrousavy Ready! No prebuilds - something is wrong with their setup in 0.82, and the app no longer builds on iOS (missing destination).

mrousavy:
Thank you!

So for testing I now have to rewrite all the tests I currently implemented in my custom testing environment, into a jest/Harness one?

If I do that, can I get all those jest/Harness tests from the app's code and run them via button clicks?

I need a single point where I define all tests, not duplicate

V3RON:
> So for testing I now have to rewrite all the tests I currently implemented in my custom testing environment, into a jest/Harness one?

Let me LLM that for you 😉 

> If I do that, can I get all those jest/Harness tests from the app's code and run them via button clicks?

What's the purpose of running tests via the UI if you already have Harness in place? 🤔

mrousavy:
> What's the purpose of running tests via the UI if you already have Harness in place? 🤔

I would say device specific issues - eg if it can only be reproduced on a real device, not with the simulator SDK. 
But this is probably not so relevant for Nitro, more so in VisionCamera.

mrousavy:
i guess i develop like this - i often run the app, run all tests, and see that something broke after my change. this allows me to iterate much faster rather than waiting for CI - but if there is a way to just run the tests locally that should be fine too

V3RON:
> > What's the purpose of running tests via the UI if you already have Harness in place? 🤔
> 
> I would say device specific issues - eg if it can only be reproduced on a real device, not with the simulator SDK. But this is probably not so relevant for Nitro, more so in VisionCamera.

No problem. You can define a Harness runner in the form of a physical device, and it will run on the device as long as the app is installed and the device is connected:
```
device: physicalAndroidDevice('Motorola', 'Moto G72'),
```

> i guess i develop like this - i often run the app, run all tests, and see that something broke after my change. this allows me to iterate much faster rather than waiting for CI - but if there is a way to just run the tests locally that should be fine too

Technically, we can wrap your tests in a factory function, function (describe, it, expect), and reuse them in both the app and Harness.

V3RON:
@mrousavy Done. All tests re-use the same definitions. You can run them via UI or via Harness.
Please approve the workflows so we can get confirmation from a CI standpoint.

mrousavy:
Hey - huge thanks for this, this is really cool.

I think I do want to simplify this a bit though considering that we have to re-implement the backend here to allow testing outside of harness, and at second thought I don't really like this - I'd like to switch it around, Harness first, and inside the example app we call the Harness test suite one test at a time - does that work?

If not, maybe it actually makes more sense to remove the tests from the example app alltogether and only run tests via CLI.

PR #1142, opened 2026-01-12
feat: Allow children on Android HybridViews

mrousavy:


mrousavy:
@hannojg we can't really do this since `ViewGroupManager<T>` requires that `T` extends `ViewGroup`.
And `View` doesn't extend `ViewGroup`. Not even `TextView` does.

PR #1149, opened 2026-01-13
chore:- expose more headers to public map

riteshshukla04:
chore:- expose more headers to public map

The build fails with if NitroModule is imported to App
image


mrousavy:
what about just `cpp/**/*.hpp` and `ios/**/*.hpp`?

riteshshukla04:
> what about just `cpp/**/*.hpp` and `ios/**/*.hpp`?

Yeah that makes sense. Done

PR #1157, opened 2026-01-15
feat: Create `SwiftConverter`

mrousavy:
## Concept

The idea is simple; currently we pass C++ STL types directly to Swift (`std::string` -> `std.string`), and also convert them inside Swift - currently this looks like this;

Swift:
```swift
func mySwiftFunc(string: std.string) {
  // 1. Convert C++ STL type (here: std::string) to Swift type (here: String)
  let swiftString = String(string)
  // 2. Call the actual Swift implementation (which obviously accepts a Swift String)
  actualSwiftImplementation.mySwiftFunc(convertedToSwift)
}
```
C++:
```cpp
void mySwiftFunc(const std::string& string) {
  // 1. Pass std::string directly to Swift
  swiftClass.mySwiftFunc(string);
}
```

To allow for extending external types, I had the idea to convert C++ STL types to Swift types _while we're still in C++ land_, so this will look like this:

Swift:
```swift
func mySwiftFunc(string: String) {
  // 1. Pass Swift String directly to our implementation
  actualSwiftImplementation.mySwiftFunc(string)
}
```
C++:
```cpp
void mySwiftFunc(const std::string& string) {
  // 1. Convert C++ STL type (here: std::string) to Swift type (here: swift::String)
  auto swiftString = SwiftConverter<std::string>::toSwift(string);
  // 2. Call the Swift method which now takes swift::String directly
  swiftClass.mySwiftFunc(swiftString);
}
```

## Implementation

As already hinted by the code above, this introduces a new template class: `SwiftConverter<T>`.

This `SwiftConverter<T>` allows to convert any C++ STL type to a Swift type and back using the Swift <> C++ interop.

This is quite tricky now, since the Swift types (`swift::String`, `swift::Optional<T>`, `swift::Array<T>`, ...) are generated **per module** (so each Pod/Framework has it's own definition of `swift::String`, even though they are identical).
I first tried to simply forward declare it, and only `NitroModules` module/Pod defines it inside it's `.cpp` file so that we don't need to define those templates multiple times - but this only works for fixed types like `swift::String`, and not for templated types like `swift::Optional<T>` or `swift::Array<T>`, since those have to be known at the time when I define the interface/hpp file.
So now it all lives inside the .hpp file.

## Progress

Currently I have implemented `SwiftConverter<T>` for `double`, `std::string`, `std::optional<T>` and `std::vector<T>`.
I am now stuck at `std::function<...>` since that requires hybrid state- one implementation has to live in Swift, one in C++, and I cannot define it just once because Swift does not have variadic generics (or at least only after iOS 17, and C++ cannot interop with that yet).
I am not sure how to implement that properly tbh.

image


mrousavy:
This is pretty cool; https://github.com/mrousavy/nitro/blob/befde766e405e2bfbb58918564990e96338bdd5e/packages/react-native-nitro-modules/ios/converter/JSIConverter%2BSwiftString.cpp#L51-L97

This probably is now the fastest way to convert a JS string to a Swift String. No intermediate C++ STL types, zero-copy segment parsing (concat is copy, but unavoidable), and this has a fast path for ASCII.

mrousavy:
Unfortunately this PR is still blocked by this Swift compiler issue: https://github.com/swiftlang/swift/issues/83373

PR #1163, opened 2026-01-16
fix:- canConvert using key instead of value

riteshshukla04:
fix:- canConvert using key instead of value


mrousavy:
Rebase with main & run `bun lint-all`

riteshshukla04:
Done

> Rebase with main & run `bun lint-all`

PR #1166, opened 2026-01-16
feat: Allow using all HybridObjects in worklets (react-native-worklets)

mrousavy:
## What

Adds support for using `HybridObject`s inside worklets by registering a custom serializable hook via a conditional import of `react-native-worklets`.

## Implementation

This essentially works by `box(...)`-ing the `HybridObject` into a `jsi::HostObject` (which react-native-worklets natively supports to serialize, since it's just a `std::shared_ptr<...>`), and then `unbox()`-ing it again in it's unpacker function.

## Discussion

This is obviously a tiny bit of an overhead compared to react-native-worklets natively using `nitro::HybridObject` APIs, but that would require react-native-worklets to depend on react-native-nitro-modules (even if it's an optional dependency, it kills their prebuild system).

So now this is JS-only, and since Nitro's `unbox()` still uses the native, fully-cached fast path (`HybridObject::toObject(..)`, which caches all prototypes) it's still super fast.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "react-native-nitro",
-  "packageManager": "bun@1.1.42",
+  "packageManager": "bun@1.3.6",
   "private": true,
   "version": "0.33.1",
   "repository": "https://github.com/mrousavy/nitro.git",


mrousavy:
Huge thanks to @tjzel and @tomekzaw for making this happen in react-native-worklets!

PR #1187, opened 2026-01-25
test: Test performance of `swift::String` direct converter

mrousavy:
Tests the performance advantages by #1157 when using Strings.

Previously (current main) we just convert a jsi::Value to a C++ STL type (for string; `std::string`), then convert that to a Swift value immediately. The C++ STL type intermediate step is kinda unnecessary, so #1157 goes the direct route; jsi::Value to Swift values directly.

The performance difference of that is pretty good, for passing a string from JS to native, it is **~2.3x faster!** 🤯 

image

The other way around (swift -> JS) is currently slower, working on it.

### DIFF
--- a/packages/react-native-nitro-test/nitrogen/generated/ios/NitroTestAutolinking.swift
+++ b/packages/react-native-nitro-test/nitrogen/generated/ios/NitroTestAutolinking.swift
@@ -36,8 +36,8 @@ public final class NitroTestAutolinking {
           return bridge.UnsafeJsiStringWrapper(consuming: facebook.jsi.String.createFromAscii(&runtime, buffer.baseAddress!, buffer.count))
         } else {
           // C) It's not ASCII, so let's use the UTF8 decoder from JSI, which internally likely transcodes to UTF16.
-          //    We cannot use a Swift fast-path here, since Swift doesn't natively stores Strings in contiguous UTF16 memory,
-          //    so this is the best we can do.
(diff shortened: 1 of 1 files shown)


mrousavy:
### UTF 8

If the string is `ascii` we take the UTF8 Path:

```swift
public static func createUTF8String(bytes: UnsafeRawPointer, length: Int) -> String {
  return String(unsafeUninitializedCapacity: length) { buffer in
    memcpy(buffer.baseAddress, bytes, length)
    return length
  }
}
```

That's very similar to C++ speed, but still 1.2x faster because of the removed conversion:

image


### UTF 16

If the string is not `ascii`, we take the UTF16 Path:

```swift
public static func createUTF16String(bytes: UnsafePointer<UInt16>, length: Int) -> String {
  return String(utf16CodeUnits: bytes, count: length)
}
```

That's much faster than the C++ path, I'm assuming because C++ does UTF16 decoding, which we then immediately do again in Swift - so now this is 2.2x faster:

image

mrousavy:
### UTF 8

Okay wow I possibly made **getting** strings insanely fast with zero-copy Swift

image

mrousavy:
### UTF 16

Now to cover all possible cases; there is a single path in which this is currently slower, and that's converting a Swift String that is not pure ASCII to JS.
The UTF16 zero-copy path is unfortunately never hit, so I have to do a copy / transcode, and that's slightly slower;

image

PR #1188, opened 2026-01-26
chore: Fix NitroExample release build after RN 0.83 upgrade on Android

riteshshukla04:
In react-native 0.83 the path to `hermesc` changed. This PR fixes the issue to update the release build path.

My android build fails with this error.
image

I think this should fix this


mrousavy:
A react-native 0.82.1 project doesn't have anything in `hermes-compiler`: 
image

riteshshukla04:
> A react-native 0.82.1 project doesn't have anything in `hermes-compiler`: image

Yeah but this is how is it in 0.83

image

PR #1189, opened 2026-01-26
fix: map `null` value to `std::nullopt` for optional types

### LINKED ISSUE #1184: Error when assigning `undefined` to optional view property
### What happened?

I have a nitro view with an optional prop. If I set that prop to a value, and then back to `undefined`, I get the error seen below.

### Reproduceable Code

Repro can be found here: https://github.com/OrfeasZ/nitro-undefined-issue

Video demonstrating the issue: https://i.nofate.me/Kj7yhtil6Ix2gAW4SKM5fl4X.mp4

Steps I took:
1. Create a sample nitro view plugin using `create-nitro-module`.
2. Make the `isRed` property optional and run `nitrogen`.
3. Adjust the relevant Swift code to handle the optional value.
4. Adjust `App.tsx` to have `isRed` set to `undefined` initially.
5. Add buttons to `App.tsx` to change `isRed` to have a value and back to `undefined`.
6. Run the example app and press on `Set Red` followed by `Clear Color`.

### Relevant log output

```shell
Exception in HostFunction: PropTest.isRed: Value is null, expected a boolean
```

### Device

iPhone 16

### Nitro Modules Version

0.33.2

### Nitrogen Version

0.33.2

I didn't try (⚠️ your issue might get ignored & closed if you don't try this)

### Additional information

- [ ] I am using Expo
- [x] I am using nitrogen

camchis:
Fixes #1184

When a prop on a fabric view is removed it's value returns `null` rather than `undefined`, this causes issues on optional types in Nitro where only `undefined` and explicitly specified values are accepted

These changes allow an optional to accept a `null` value which will be mapped to `std::nullopt`

If an optional has `null` specified as an accepted type then this will not be affected by these changes and you will still have distinction between `null` and `undefined` values if needed

### DIFF
--- a/packages/react-native-nitro-modules/cpp/jsi/JSIConverter+Optional.hpp
+++ b/packages/react-native-nitro-modules/cpp/jsi/JSIConverter+Optional.hpp
@@ -25,9 +25,14 @@ struct JSIConverter<std::optional<TInner>> final {
   static inline std::optional<TInner> fromJSI(jsi::Runtime& runtime, const jsi::Value& arg) {
     if (arg.isUndefined()) {
       return std::nullopt;
-    } else {
-      return JSIConverter<TInner>::fromJSI(runtime, arg);
     }
+    if (arg.isNull()) {
+      if (JSIConverter<TInner>::canConvert(runtime, arg)) {
+        return JSIConverter<TInner>::fromJSI(runtime, arg);
+      }
+      return std::nullopt;
+    }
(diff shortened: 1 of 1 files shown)


mrousavy:
I'm not sure if this is a good idea tbh. This loosens null/undefined conversions, which in JS are two separate concepts.

I think we should configure our React components to give us `undefined` by default instead of `null` for props. cc @hannojg

PR #1190, opened 2026-01-27
perf:- preallocate size during AnyMap::make() . ~12% faster

riteshshukla04:
image
image

To be honest I was lucky here to get this difference (when I took this screenshot) . After that I tried few more times. This was around 12%-15%faster


mrousavy:
Wait so is the difference you showed an extreme example, or is this actually the average?

PR #1196, opened 2026-02-03
Update import statement for NitroModulesSpec

WoLewicki:
The import should be done in `<>` ticks since it is not local, its from ReactCodegen pod. See the same e.g. here: https://github.com/software-mansion/react-native-gesture-handler/blob/9f573ef7bbfcd6445822bf6bb8bfd3f4ddce1c93/packages/react-native-gesture-handler/apple/RNGestureHandlerModule.h#L7. cc @hannojg


mrousavy:
Dziękuję !

PR #1289, opened 2026-04-15
fix: Fix case where a struct's `operator==` symbol cannot be linked to from Swift (`Equatable` conformance)

mrousavy:
Tries to reproduce & fix the error from https://github.com/mrousavy/react-native-vision-camera/blob/7586e98f53bea81db67035d7534da0249b4f7511/packages/react-native-vision-camera/ios/Hybrid%20Objects/Inputs/HybridCameraDevice.swift#L58-L68


mrousavy:
hm, I cannot reproduce [the build error from VisionCamera V5](https://github.com/mrousavy/react-native-vision-camera/blob/7586e98f53bea81db67035d7534da0249b4f7511/packages/react-native-vision-camera/ios/Hybrid%20Objects/Inputs/HybridCameraDevice.swift#L58-L68) here now.

PR #1291, opened 2026-04-18
fix(docs): update docs to reflect new autolinking syntax

DaveyEke:
This PR updates [this part of the docs](https://nitro.margelo.com/docs/guides/view-components#4-autolink) to relfect the new autolinking syntax


mrousavy:
thanks!

Issue #1319, opened 2026-05-08
Optional<Struct> reads garbage enum tag bytes on iOS Release via Swift Cxx-interop bridge


notsuhas:
### What happened?

`Optional<Struct>` round-trips through the Swift Cxx-interop bridge corrupt enum tag bytes when the wrapping struct is wide and contains multiple distinct string-union enum optional fields. The C++-side JS converter then sees a non-discriminator integer in the enum slot and throws e.g. `Cannot convert Tier to JS - invalid value: 83886081`.

The garbage value differs per call (observed `83886081` and `2003789939` in separate runs), and the bug surfaces only under optimization (Swift `-O` / Clang `-Os`/`-O3`). iOS Debug, Android (no Cxx-interop in play), and narrower struct shapes all round-trip cleanly.

#### Surface conditions

| Build         | Result                                                                          |
|---------------|---------------------------------------------------------------------------------|
| iOS Debug     | ✅ passes                                                                        |
| iOS Release   | ❌ `tryOptionalEnumStruct(...) equals` fails — `Cannot convert <Enum> to JS - invalid value: <garbage int>` |
| Android       | ✅ passes on both `TestObjectCpp` and `TestObjectSwiftKotlin` (no Cxx-interop)   |

#### Bisection (three repro tests on `TestObject`, increasing in width)

| Repro                                              | Shape                                                                        | iOS Release |
|----------------------------------------------------|------------------------------------------------------------------------------|:-----------:|
| `tryOptionalStruct(value?: OptionalWrapper)`       | 2 primitive optionals, no enums                                              | ✅          |
| `tryOptionalNestedStruct(value?: OptionalNested…)` | 3 fields incl. nested struct with 2× same `Side` enum                         | ✅          |
| `tryOptionalEnumStruct(value?: OptionalEnum…)`     | 14 primitive optionals + 2 distinct enums + nested struct with 2 more enums | ❌          |

Together these bracket the bug: `small + few enums = works`, `wide + multiple distinct string-union enum types under -O = corrupted enum tag`. Garbage values being non-deterministic across calls (and different from a stable wrong-pointer value) point at uninitialized padding bytes or an optimizer-revealed layout mismatch in Swift's `Optional<Struct>` projection across the Cxx-interop bridge — not a stable wrong-pointer or logic bug.

### Reproduceable Code

Repro branch (on top of `mrousavy/nitro@main`): https://github.com/notsuhas/nitro/tree/repro/optional-struct-bridge

[...]

PR #1320, opened 2026-05-08
test: repro Optional<Struct> garbage tag read on Swift Cxx-interop

notsuhas:
Repro-only PR for #1319, per the [repro-only flow](https://github.com/mrousavy/nitro/blob/main/CONTRIBUTING.md#you-dont-need-to-ship-a-fix--a-clean-repro-is-enough). **Harness-iOS Release is expected to fail** on `tryOptionalEnumStruct(...) equals` — that failure is the point of this PR.

## Repro

Adds **1 new struct** (`OptionalEnumWrapper`) and **1 new method** (`tryOptionalEnumStruct`) on `TestObject`. No new enums — reuses the existing `Powertrain` for the four enum fields. The wrapper mixes 14 primitive optionals (numbers + booleans) with 4 string-union-enum optionals.

Native impls are identity bounces (1 line each on Cpp / Swift / Kotlin). Rest of the diff is nitrogen-regenerated per CONTRIBUTING.md.

## Failure surface

iOS Release on a real device: 240/241 tests pass. `tryOptionalEnumStruct(...) equals` fails with number fields read as `5e-324` / `2.121995791e-314` — i.e. subnormal denormalized doubles, bit pattern `0x0000000000000001`. Classic garbage-tag read out of `std::optional<Struct>` under `-O` on Swift Cxx-interop. The expected fields like `count` drop entirely (the optional tag bits read garbage → "not present"). iOS Debug + C++ + Android all pass cleanly.

Likely root cause: `.value` direct projection on `std::optional<T>` is fragile when the inner `T` is a struct mixing bool/number/string-union-enum fields. Full diagnostic notes in #1319.

## History

Original PR added 9 types + 3 methods. Slimmed in `60d2f03` per @mrousavy's review — only `tryOptionalEnumStruct` + `OptionalEnumWrapper` reproduce the bug; the narrower `tryOptionalStruct`/`tryOptionalNestedStruct` cases pass cleanly and were dropped along with the 7 supporting types.


notsuhas:
Heads up on CI coverage for this repro: `harness-ios.yml` builds with `-configuration Debug` ([line 141](https://github.com/mrousavy/nitro/blob/main/.github/workflows/harness-ios.yml#L141)), and the bug only surfaces under Swift `-O` (iOS Release). iOS Debug uses `-Onone` and passes the new `tryOptionalEnumStruct(...) equals` test, so Harness-iOS will go green here despite the regression being real.

`build-ios-release.yml` does build Release but is compile-only and only triggers on `release: published` / when the workflow file itself changes — it won't catch a runtime-Release bug like this either.

Verified the failure on-device against an iOS Release build of the example app (full diagnostic notes in #1319).

mrousavy (changes requested):
Hey - first of all, thank you for the repro for this - this is how we develop Nitro and is generally the right direction.

However, as per [Nitro's `CONTRIBUTING.md`](https://github.com/mrousavy/nitro/blob/main/CONTRIBUTING.md), we need to keep those repros to be absolutely as minimal as possible, and your repro introduces **9 new types** and 3 new methods.
This will make the test app/codebase huge as we add more and more repros, which increases compilation time and is generally harder to keep track of.

It seems like this was generated with Claude, so please prompt Claude to strip this repro down to the absolute minimum code here to reproduce this, ideally by also re-using the types we already have here and only creating 1 or 2 new types max.

Let me know if this is not possible and you really need all 9 types and 3 methods.

notsuhas:
Slimmed the repro per the review — pushed as `60d2f03`.

**Diff:** 1 new struct (`OptionalEnumWrapper`), 1 new method (`tryOptionalEnumStruct`), 0 new enums (reuses existing `Powertrain`). Dropped `tryOptionalStruct` + `tryOptionalNestedStruct` and the `Side`/`Stage`/`Tier`/`Region`/`Tone`/`OptionalNestedInner`/`OptionalNestedWrapper`/`OptionalEnumInner` types. Inlined `OptionalEnumInner`'s fields into the wrapper.

Verified the bug still reproduces on a real iPhone in Release: 240/241 tests pass, only `tryOptionalEnumStruct(...) equals` fails on the Swift target. Number fields come back as `5e-324` / `2.12e-314` — subnormal denorms, classic garbage-tag read on `std::optional<Struct>` under -O. C++ target passes, as expected.

Kept the wide field mix (~14 primitives + 4 enums) because trimming further made the bug stop reproducing in earlier experiments — the workaround in `SwiftCxxBridgedType.ts` is fragile specifically when many bool/number/enum fields share an `optional<Struct>`. Happy to push further if you have a tighter shape in mind.

mrousavy:
Ok and the CI is expected to go green here, but fail in release right?

notsuhas:
@mrousavy Yes, as the iOS harness is running in debug, it does not reproduce this failure. Maybe a separate workflow that runs this on release. Maybe that should work? Let me know; maybe I can try it out.

mrousavy:
Yep no that's fine - we can reproduce & test this here, and manually run release tests once in a while - you can run the release app which is being built & attached to every GitHub release manually.

PR #1326, opened 2026-05-19
fix: Reset Kotlin C++ state in `dispose()`

mrousavy:
`dispose()` calls into the platform implemented `dispose()` method (C++, Swift or Kotlin), and then unlinks the `jsi::NativeState` from the `jsi::Object`.

This works fine for ref-counted languages like C++ or Swift, as if no more references are being held, it deletes all associated objects pretty much instantly.

For Kotlin, this is a bit different. The Kotlin HybridObject is garbage-collected, but its `CxxPart` also owns a `HybridData`, and fbjni keeps a JNI `global_ref` around for the Java CxxPart.

So the structure is roughly like this:

```txt
Java/Kotlin CxxPart
  ├── javaPart: Kotlin HybridObject
  └── mHybridData: HybridData
          │
          ▼
C++ JHybridX::CxxPart
  └── _cxxJavaPart: global_ref<Java/Kotlin CxxPart>
          │
          └──────────────▶ Java/Kotlin CxxPart
```

If we only dispose the JS/C++ HybridObject state, the Kotlin/Java CxxPart can stay around and keep JNI global refs alive. After enough manual `dispose()` calls this can crash with:

```txt
JNI ERROR (app bug): global reference table overflow (max=51200)
```

Reported in:
- https://github.com/mrousavy/react-native-vision-camera/issues/3879
- https://github.com/mrousavy/react-native-vision-camera/issues/3824#issuecomment-4475912523

To fix this, Kotlin `HybridObject.dispose()` now eagerly destroys the connected `CxxPart` by calling `HybridData.resetNative()`, and then clears the cached CxxPart reference.

I also added `@CallSuper` to `dispose()`, made `CxxPart.destroy()` non-overridable, and fixed our own Kotlin test object to call `super.dispose()`.

The Harness now also has a manual `dispose()` memory test that creates/disposes 55k HybridObjects, because JNI's global ref limit is 51200. The old Harness test only covered automatic GC cleanup, not this repeated manual `dispose()` path.

### DIFF
--- a/example/src/getTests.ts
+++ b/example/src/getTests.ts
@@ -1537,9 +1537,11 @@ export function getTests(
         const baselineAllocations =
           NitroModules.debug_getTotalAllocatedHybridObjects()
 
+        // We test 51_201, because JNI's max global_ref allocations is 51200,
+        // so if this test goes green, it properly deletes jni::global_refs.
+        // If there would be a memory leak, this test will likely crash in C++.
+        const TOTAL_ALLOCATIONS = 51_201
         const BATCH_SIZE = 1000
-        const LOOP_COUNT = 10
-        const TOTAL_ALLOCATIONS = BATCH_SIZE * LOOP_COUNT
 
         const objects: Array<TestObjectCpp | TestObjectSwiftKotlin> = []
         for (let i = 0; i < TOTAL_ALLOCATIONS; i++) {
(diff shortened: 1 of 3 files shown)


mrousavy:
Managed to reproduce it: without this fix it goes red - so this PR actually fixed it! https://github.com/mrousavy/nitro/actions/runs/26094138868/job/76727279854?pr=1329

PR #1332, opened 2026-05-19
fix: Check generated hybrid bridge casts

mrousavy:


mrousavy:
not needed i mean we know what type it is

PR #1358, opened 2026-05-27
chore: Upgrade react-native-builder-bob to 0.41.0

mrousavy:
## Summary

- upgrade `react-native-builder-bob` from `0.37.0` to `0.41.0`
- refresh the Bun lockfile for Bob's updated build dependency graph

## Notes

Bob 0.40.0 changed the default template away from the previous dual CJS/ESM package setup toward ESM-only output. This package currently publishes both `main` and `module` outputs, and changing that would be a consumer-facing module-format migration, so this PR keeps the existing Bob targets intact.

Bob 0.41.0 also introduces a stricter `tsc` lookup path. In this Bun workspace it builds successfully by falling back to the repo-root `node_modules/.bin/tsc`, but prints a warning because Bob checks for `node_modules/.bin/tsc` under the package directory first. I did not hardcode a repo-relative `tsc` path into the published package config.

## Validation

- `bun install --ignore-scripts`
- `bun nitro typecheck`
- `bun nitro build`
- `git diff --check`

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -115,7 +115,7 @@
         "jest": "*",
         "react": "19.2.6",
         "react-native": "0.85.3",
-        "react-native-builder-bob": "^0.37.0",
+        "react-native-builder-bob": "^0.41.0",
         "react-native-worklets": "^0.9.1",
       },
       "peerDependencies": {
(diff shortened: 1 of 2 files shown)


mrousavy:
Superseded by #1362, which folds the remaining package upgrades into a single PR.

PR #1397, opened 2026-06-19
test(nitro-test): reproduce C++ Promise<T> data race (resolve vs JSIConverter::toJSI)

mfazekas:
Reproduction PR for #1396 (per the [contributing guide](https://nitro.margelo.com/docs/contributing)).

## What this adds
- `TestObjectCpp.newTestObjectAsync()` — returns a HybridObject from a `Promise::async`, i.e. the Promise is **resolved off the JS thread** (on the Nitro ThreadPool).
- A `getTests.ts` test that hammers it under parallel load (`20 × 32` concurrent calls).

## The race
Between `Promise<T>::resolve()` — which writes `_state` under `_mutex` — and `JSIConverter<shared_ptr<Promise<T>>>::toJSI`, which reads `_state` **lock-free** via `isPending()`/`isResolved()`/`getResult()` on the JS thread the instant the hybrid method returns. No happens-before → data race on the `std::variant _state`.

## How to see it
The test **passes functionally** — the verdict is Thread Sanitizer. Enable it from Xcode:

1. Open `example/ios/NitroExample.xcworkspace`.
2. **Product → Scheme → Edit Scheme…** (`⌘<`) → **Run** → **Diagnostics** tab → check **Thread Sanitizer**.
3. Pick an iOS Simulator and **Run** (`⌘R`).
4. On the **HybridObject Tests** screen, run the `newTestObjectAsync() under parallel load …` test (or **Run all tests**).

TSan reports the race on `Promise::_state` (`resolve` at `Promise.hpp:105` vs `isPending` at `Promise.hpp:248`) in the debug console / Issue Navigator. The test itself still shows ✅ — TSan is the signal. Confirmed on iPhone 17 simulator (iOS 26.4).

image

My debug log looks like this:

```log
[HybridObjectRegistry] Registering HybridObject "TestObjectCpp"...
[HybridObjectRegistry] Successfully registered HybridObject "TestObjectCpp"!
[HybridObjectRegistry] Registering HybridObject "TestObjectSwiftKotlin"...
[HybridObjectRegistry] Successfully registered HybridObject "TestObjectSwiftKotlin"!
[HybridObjectRegistry] Registering HybridObject "Base"...
[HybridObjectRegistry] Successfully registered HybridObject "Base"!
[HybridObjectRegistry] Registering HybridObject "Child"...
[HybridObjectRegistry] Successfully registered HybridObject "Child"!
[HybridObjectRegistry] Registering HybridObject "PlatformObject"...
[...]

PR #1398, opened 2026-06-19
fix(core): synchronize C++ Promise<T> state-inspection accessors

### LINKED ISSUE #1396: Data race in C++ `Promise<T>`: `_state` read lock-free by `toJSI` while `resolve()` writes under `_mutex`
### What happened?

We had some hard-to-reproduce [crashes](https://github.com/rive-app/rive-nitro-react-native/issues/297) in nitro-rive. We could reproduce some of the issue with Thread Sanitizer in Rive-specific code, but it seems Nitro itself has a data race as well.

Context: the Rive runtime must be touched only on the **main thread**, so in our hybrid methods we hop the work to the main thread and call `Promise.resolve()` there — i.e. the Promise is resolved off the JS thread.

TSan detects occasional data races between the promise [`resolve()`](https://github.com/mrousavy/nitro/blob/334f8a9f6bb4d32a285dc5f09f0625d8ff67ceed/packages/react-native-nitro-modules/cpp/core/Promise.hpp#L100-L110) on the main thread and [`JSIConverter<shared_ptr<Promise<T>>>::toJSI`](https://github.com/mrousavy/nitro/blob/334f8a9f6bb4d32a285dc5f09f0625d8ff67ceed/packages/react-native-nitro-modules/cpp/jsi/JSIConverter+Promise.hpp#L55-L95) on the JS thread. 

Note: `resolve()` writes the [`_state`](https://github.com/mrousavy/nitro/blob/334f8a9f6bb4d32a285dc5f09f0625d8ff67ceed/packages/react-native-nitro-modules/cpp/core/Promise.hpp#L258-L261) variant under `_mutex`, but the reads `toJSI` uses — [`isPending()`](https://github.com/mrousavy/nitro/blob/334f8a9f6bb4d32a285dc5f09f0625d8ff67ceed/packages/react-native-nitro-modules/cpp/core/Promise.hpp#L246-L249) / [`isResolved()`](https://github.com/mrousavy/nitro/blob/334f8a9f6bb4d32a285dc5f09f0625d8ff67ceed/packages/react-native-nitro-modules/cpp/core/Promise.hpp#L232-L235) / [`getResult()`](https://github.com/mrousavy/nitro/blob/334f8a9f6bb4d32a285dc5f09f0625d8ff67ceed/packages/react-native-nitro-modules/cpp/core/Promise.hpp#L208-L215) — take no lock, so there's no happens-before between the two threads.

Worst case, `toJSI` reads the variant's tag as `resolved` before the payload store is visible, then `getResult()` hands JSI a **torn `shared_ptr<HybridObject>`** → crash / heap corruption.

### Reproduceable Code

```tsx
Added to `nitro-test` `TestObjectCpp` — `newTestObjectAsync()` next to the existing [`newTestObject()`](https://github.com/mrousavy/nitro/blob/334f8a9f6bb4d32a285dc5f09f0625d8ff67ceed/packages/react-native-nitro-test/src/specs/TestObject.nitro.ts#L344) ([impl here](https://github.com/mrousavy/nitro/blob/334f8a9f6bb4d32a285dc5f09f0625d8ff67ceed/packages/react-native-nitro-test/cpp/HybridTestObjectCpp.cpp#L689)), and the harness next to [`example/__tests__/nitro.harness.ts`](https://github.com/mrousavy/nitro/blob/334f8a9f6bb4d32a285dc5f09f0625d8ff67ceed/example/__tests__/nitro.harness.ts):


// HybridTestObjectCpp.cpp — resolves a HybridObject off the JS thread
std::shared_ptr<Promise<std::shared_ptr<HybridTestObjectCppSpec>>> HybridTestObjectCpp::newTestObjectAsync() {
  return Promise<std::shared_ptr<HybridTestObjectCppSpec>>::async(
      []() -> std::shared_ptr<HybridTestObjectCppSpec> { return std::make_shared<HybridTestObjectCpp>(); });
}


// harness — hammer it
for (let r = 0; r < 100; r++) await Promise.all(Array.from({ length: 32 }, () => obj.newTestObjectAsync()))


> ℹ️ Build the example with the `ENABLE_THREAD_SANITIZER=YES` build setting, and confirm the binary is actually instrumented:
> 
> nm -u NitroExample.app/NitroExample.debug.dylib | grep __tsan_func_entry   # non-empty ⇒ code is TSan-instrumented
> otool -L NitroExample.app/NitroExample | grep tsan                         # ⇒ libclang_rt.tsan_*.dylib linked
> 
> (Release build: the code is in `NitroExample.app/NitroExample` rather than the `.debug.dylib`.)
```

### Relevant log output

```shell
WARNING: ThreadSanitizer: data race
  Write of size 4 by thread T21 (mutexes: write M0):                 // resolve() on the ThreadPool, holds _mutex
    Promise<shared_ptr<HybridTestObjectCppSpec>>::resolve(…)  Promise.hpp:105
  Previous read of size 4 by thread T7:                              // JS thread, toJSI — NO mutex
    Promise<shared_ptr<HybridTestObjectCppSpec>>::isPending() const  Promise.hpp:248
  Mutex M0 created at:  Promise<…>::resolve(…)  Promise.hpp:101
SUMMARY: ThreadSanitizer: data race in std::variant<…HybridTestObjectCppSpec…>  (Promise::_state)
```

### Device

iPhone 17 simulator (iOS 26.4); the defect is pure C++ so iOS + Android are equally affected.

### Nitro Modules Version

0.35.9 (the C++ `_mutex` exists since #439 but never covered the inspection accessors; distinct from the Swift-layer fix in #1338).

### Nitrogen Version

_No response_

### Additional information

- [ ] I am using Expo
- [x] I am using nitrogen
- [x] I created a reproduction PR #1397 to reproduce this issue here in the nitro repo. (See [Contributing](https://nitro.margelo.com/docs/contributing) for more information)

mfazekas:
Fixes #1396.

`Promise<T>::resolve()`/`reject()` write `_state` under `_mutex`, but the state-inspection accessors (`isPending`/`isResolved`/`isRejected`/`getResult`/`getError`) read it **lock-free**. `JSIConverter<shared_ptr<Promise<T>>>::toJSI` runs those reads on the JS thread the instant a hybrid method returns, racing an off-JS-thread `resolve()` (`Promise::async` thread pool, `Promise.parallel`, or a main-thread resolve) — a data race on the `std::variant _state`.

## Fix
- Make `_mutex` `mutable` and take it in the inspection accessors (state check inlined to avoid recursive self-locking).
- Per-accessor locking is sufficient — no combined snapshot needed: `_state`'s transition is one-shot/monotonic (`pending → value | error`, never back), so the `isPending()` → `getResult()` check-then-act is safe, and `getResult()` may still return a reference (the value is immutable after resolution; the lock just supplies the happens-before).
- Pop listeners under the lock and invoke them outside it (matching the Swift fix in #1338), so user callbacks never run while holding `_mutex`. Same treatment for the `Promise<void>` specialization.

## Verification
Reproduction in #1397 (adds `newTestObjectAsync` + a parallel-load test). Building the example with Thread Sanitizer (`ENABLE_THREAD_SANITIZER=YES` / scheme Diagnostics):
- **Before:** TSan reports the race on `Promise::_state` (`resolve` at `Promise.hpp:105` vs `isPending` at `Promise.hpp:248`, via `toJSI`).
- **After:** that race is gone. Confirmed on iPhone 17 simulator (iOS 26.4), and via a standalone TSan harness against the real `Promise.hpp` (0 races, was ~4/run).

> Out of scope: the same TSan run also surfaces a separate, pre-existing `ThreadPool::run()` deque race — not addressed here.

### DIFF
--- a/packages/react-native-nitro-modules/cpp/core/Promise.cpp
+++ b/packages/react-native-nitro-modules/cpp/core/Promise.cpp
@@ -56,11 +56,12 @@ void Promise<void>::resolve() {
   assertPromiseState(*this, PromiseTask::WANTS_TO_RESOLVE);
 #endif
   std::vector<OnResolvedFunc> listeners;
+  std::vector<OnRejectedFunc> dropped;
   {
     std::unique_lock lock(_mutex);
     _isResolved = true;
     listeners = std::move(_onResolvedListeners);
-    clearAllListeners(lock);
+    dropped = std::move(_onRejectedListeners);
   }
   for (const auto& onResolved : listeners) {
     onResolved();
@@ -76,11 +77,12 @@ void Promise<void>::reject(const std::exception_ptr& exception) {
   assertPromiseState(*this, PromiseTask::WANTS_TO_REJECT);
 #endif
   std::vector<OnRejectedFunc> listeners;
+  std::vector<OnResolvedFunc> dropped;
   {
     std::unique_lock lock(_mutex);
     _error = exception;
     listeners = std::move(_onRejectedListeners);
-    clearAllListeners(lock);
+    dropped = std::move(_onResolvedListeners);
   }
   for (const auto& onRejected : listeners) {
     onRejected(exception);
@@ -147,11 +149,6 @@ const std::exception_ptr& Promise<void>::getError() const {
     throw std::runtime_error("Cannot get error when Promise<void> is not yet rejected!");
   }
   return _error;
-}
-
-void Promise<void>::clearAllListeners(const std::unique_lock<std::mutex>&) noexcept {
-  _onResolvedListeners.clear();
-  _onRejectedListeners.clear();
 }
 
 } // namespace margelo::nitro
--- a/packages/react-native-nitro-modules/cpp/core/Promise.cpp
+++ b/packages/react-native-nitro-modules/cpp/core/Promise.cpp
@@ -121,7 +121,6 @@ void Promise<void>::addOnRejectedListener(OnRejectedFunc&& onRejected) {
 void Promise<void>::addOnRejectedListener(const OnRejectedFunc& onRejected) {
   std::unique_lock lock(_mutex);
   if (isRejected(lock)) {
-    // Promise is already rejected! Call the callback immediately.
     std::exception_ptr error = _error;
     lock.unlock();
     onRejected(error);
--- a/packages/react-native-nitro-modules/cpp/core/Promise.cpp
+++ b/packages/react-native-nitro-modules/cpp/core/Promise.cpp
@@ -92,29 +92,24 @@ void Promise<void>::reject(const std::exception_ptr& exception) {
 void Promise<void>::addOnResolvedListener(OnResolvedFunc&& onResolved) {
   std::unique_lock lock(_mutex);
   if (isResolved(lock)) {
-    // Promise is already resolved! Call the callback immediately.
     lock.unlock();
     onResolved();
   } else {
-    // Promise is not yet resolved, put the listener in our queue.
     _onResolvedListeners.push_back(std::move(onResolved));
   }
 }
 void Promise<void>::addOnResolvedListener(const OnResolvedFunc& onResolved) {
   std::unique_lock lock(_mutex);
   if (isResolved(lock)) {
-    // Promise is already resolved! Call the callback immediately.
     lock.unlock();
     onResolved();
   } else {
-    // Promise is not yet resolved, put the listener in our queue.
     _onResolvedListeners.push_back(onResolved);
   }
 }
 void Promise<void>::addOnRejectedListener(OnRejectedFunc&& onRejected) {
   std::unique_lock lock(_mutex);
   if (isRejected(lock)) {
-    // Promise is already rejected! Call the callback immediately.
     std::exception_ptr error = _error;
     lock.unlock();
     onRejected(error);
--- a/packages/react-native-nitro-modules/cpp/core/Promise.cpp
+++ b/packages/react-native-nitro-modules/cpp/core/Promise.cpp
@@ -59,7 +59,8 @@ void Promise<void>::resolve() {
   {
     std::unique_lock lock(_mutex);
     _isResolved = true;
-    listeners = takeResolvedListenersAndClearAllListeners(lock);
+    listeners = std::move(_onResolvedListeners);
+    clearAllListeners(lock);
   }
   for (const auto& onResolved : listeners) {
     onResolved();
@@ -78,7 +79,8 @@ void Promise<void>::reject(const std::exception_ptr& exception) {
   {
     std::unique_lock lock(_mutex);
     _error = exception;
-    listeners = takeRejectedListenersAndClearAllListeners(lock);
+    listeners = std::move(_onRejectedListeners);
+    clearAllListeners(lock);
   }
   for (const auto& onRejected : listeners) {
     onRejected(exception);
@@ -86,48 +88,50 @@ void Promise<void>::reject(const std::exception_ptr& exception) {
 }
 
 void Promise<void>::addOnResolvedListener(OnResolvedFunc&& onResolved) {
-  {
-    std::unique_lock lock(_mutex);
-    if (!isResolved(lock)) {
-      _onResolvedListeners.push_back(std::move(onResolved));
-      return;
-    }
+  std::unique_lock lock(_mutex);
+  if (isResolved(lock)) {
+    // Promise is already resolved! Call the callback immediately.
+    lock.unlock();
+    onResolved();
+  } else {
+    // Promise is not yet resolved, put the listener in our queue.
+    _onResolvedListeners.push_back(std::move(onResolved));
   }
-  onResolved();
 }
 void Promise<void>::addOnResolvedListener(const OnResolvedFunc& onResolved) {
-  {
-    std::unique_lock lock(_mutex);
-    if (!isResolved(lock)) {
-      _onResolvedListeners.push_back(onResolved);
-      return;
-    }
+  std::unique_lock lock(_mutex);
+  if (isResolved(lock)) {
+    // Promise is already resolved! Call the callback immediately.
+    lock.unlock();
+    onResolved();
+  } else {
+    // Promise is not yet resolved, put the listener in our queue.
+    _onResolvedListeners.push_back(onResolved);
   }
-  onResolved();
 }
 void Promise<void>::addOnRejectedListener(OnRejectedFunc&& onRejected) {
-  std::exception_ptr error;
-  {
-    std::unique_lock lock(_mutex);
-    if (!isRejected(lock)) {
-      _onRejectedListeners.push_back(std::move(onRejected));
-      return;
-    }
-    error = _error;
+  std::unique_lock lock(_mutex);
+  if (isRejected(lock)) {
+    // Promise is already rejected! Call the callback immediately.
+    std::exception_ptr error = _error;
+    lock.unlock();
+    onRejected(error);
+  } else {
+    // Promise is not yet rejected, put the listener in our queue.
+    _onRejectedListeners.push_back(std::move(onRejected));
   }
-  onRejected(error);
 }
 void Promise<void>::addOnRejectedListener(const OnRejectedFunc& onRejected) {
-  std::exception_ptr error;
-  {
-    std::unique_lock lock(_mutex);
-    if (!isRejected(lock)) {
-      _onRejectedListeners.push_back(onRejected);
-      return;
-    }
-    error = _error;
+  std::unique_lock lock(_mutex);
+  if (isRejected(lock)) {
+    // Promise is already rejected! Call the callback immediately.
+    std::exception_ptr error = _error;
+    lock.unlock();
+    onRejected(error);
+  } else {
+    // Promise is not yet rejected, put the listener in our queue.
+    _onRejectedListeners.push_back(onRejected);
   }
-  onRejected(error);
 }
 
 std::future<void> Promise<void>::await() {
@@ -143,18 +147,6 @@ const std::exception_ptr& Promise<void>::getError() const {
     throw std::runtime_error("Cannot get error when Promise<void> is not yet rejected!");
   }
   return _error;
-}
-
-std::vector<Promise<void>::OnResolvedFunc> Promise<void>::takeResolvedListenersAndClearAllListeners(const std::unique_lock<std::mutex>& lock) {
-  auto listeners = std::move(_onResolvedListeners);
-  clearAllListeners(lock);
-  return listeners;
-}
-
-std::vector<Promise<void>::OnRejectedFunc> Promise<void>::takeRejectedListenersAndClearAllListeners(const std::unique_lock<std::mutex>& lock) {
-  auto listeners = std::move(_onRejectedListeners);
-  clearAllListeners(lock);
-  return listeners;
 }
 
 void Promise<void>::clearAllListeners(const std::unique_lock<std::mutex>&) noexcept {
--- a/packages/react-native-nitro-modules/cpp/core/Promise.cpp
+++ b/packages/react-native-nitro-modules/cpp/core/Promise.cpp
@@ -52,15 +52,18 @@ std::shared_ptr<Promise<void>> Promise<void>::rejected(const std::exception_ptr&
 }
 
 void Promise<void>::resolve() {
-  std::unique_lock lock(_mutex);
 #ifdef NITRO_DEBUG
   assertPromiseState(*this, PromiseTask::WANTS_TO_RESOLVE);
 #endif
-  _isResolved = true;
-  for (const auto& onResolved : _onResolvedListeners) {
+  std::vector<OnResolvedFunc> listeners;
+  {
+    std::unique_lock lock(_mutex);
+    _isResolved = true;
+    listeners = takeResolvedListenersAndClearAllListeners(lock);
+  }
+  for (const auto& onResolved : listeners) {
     onResolved();
   }
-  didFinish();
 }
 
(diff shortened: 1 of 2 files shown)


packages/react-native-nitro-modules/cpp/core/Promise.cpp
```diff
@@ -52,66 +52,82 @@ std::shared_ptr<Promise<void>> Promise<void>::rejected(const std::exception_ptr&
...
 #ifdef NITRO_DEBUG
   assertPromiseState(*this, PromiseTask::WANTS_TO_REJECT);
 #endif
-  _error = exception;
-  for (const auto& onRejected : _onRejectedListeners) {
+  std::vector<OnRejectedFunc> listeners;
+  {
+    std::unique_lock lock(_mutex);
+    _error = exception;
+    listeners = takeRejectedListenersAndClearAllListeners(lock);
```
mrousavy: Please remove `takeRejectedListenersAndClearAllListeners` - this adds more lines than it eliminates. This would be a simple `std::move(_onRejectedListeners);` + `clearAllListeners(lock)`.

packages/react-native-nitro-modules/cpp/core/Promise.cpp
```diff
@@ -52,66 +52,82 @@ std::shared_ptr<Promise<void>> Promise<void>::rejected(const std::exception_ptr&
...
 #ifdef NITRO_DEBUG
   assertPromiseState(*this, PromiseTask::WANTS_TO_RESOLVE);
 #endif
-  _isResolved = true;
-  for (const auto& onResolved : _onResolvedListeners) {
+  std::vector<OnResolvedFunc> listeners;
+  {
+    std::unique_lock lock(_mutex);
+    _isResolved = true;
+    listeners = takeResolvedListenersAndClearAllListeners(lock);
```
mrousavy: Please remove `takeResolvedListenersAndClearAllListeners` - this adds more lines than it eliminates. This would be a simple `std::move(_onResolvedListeners);` + `clearAllListeners(lock)`.

packages/react-native-nitro-modules/cpp/core/Promise.hpp
```diff
@@ -98,26 +98,40 @@ class Promise final {
...
     assertPromiseState(*this, PromiseTask::WANTS_TO_RESOLVE);
 #endif
-    _state = std::move(result);
-    for (const auto& onResolved : _onResolvedListeners) {
-      onResolved(std::get<TResult>(_state));
+    std::vector<OnResolvedFunc> listeners;
+    {
+      std::unique_lock lock(_mutex);
+      _state = std::move(result);
+      listeners = takeResolvedListenersAndClearAllListeners(lock);
```
mrousavy: same here

packages/react-native-nitro-modules/cpp/core/Promise.hpp
```diff
@@ -98,26 +98,40 @@ class Promise final {
...
     assertPromiseState(*this, PromiseTask::WANTS_TO_RESOLVE);
 #endif
-    _state = result;
-    for (const auto& onResolved : _onResolvedListeners) {
-      onResolved(std::get<TResult>(_state));
+    std::vector<OnResolvedFunc> listeners;
+    {
+      std::unique_lock lock(_mutex);
+      _state = result;
+      listeners = takeResolvedListenersAndClearAllListeners(lock);
```
mrousavy: same here

packages/react-native-nitro-modules/cpp/core/Promise.hpp
```diff
@@ -128,15 +142,18 @@ class Promise final {
...
 #ifdef NITRO_DEBUG
     assertPromiseState(*this, PromiseTask::WANTS_TO_REJECT);
 #endif
-    _state = exception;
-    for (const auto& onRejected : _onRejectedListeners) {
+    std::vector<OnRejectedFunc> listeners;
+    {
+      std::unique_lock lock(_mutex);
+      _state = exception;
+      listeners = takeRejectedListenersAndClearAllListeners(lock);
```
mrousavy: same here

packages/react-native-nitro-modules/cpp/core/Promise.hpp
```diff
@@ -145,49 +162,53 @@ class Promise final {
...
-      // Promise is not yet resolved, put the listener in our queue.
-      _onResolvedListeners.push_back(std::move(onResolved));
+    {
+      std::unique_lock lock(_mutex);
+      if (!isResolved(lock)) {
+        _onResolvedListeners.push_back(std::move(onResolved));
+        return;
+      }
     }
+    onResolved(getResult());
```
mrousavy: I don't like this diff. It reads much harder than the part before - before it was lock, check if it is resolved, if yes call callback now, if not add listener.
Fast path first, longer path fallback. Easy to understand. No early return, no messy RAII blocks, no default path always being hit except for early returns, no nothing.

packages/react-native-nitro-modules/cpp/core/Promise.hpp
```diff
@@ -145,49 +162,53 @@ class Promise final {
...
+    std::exception_ptr error;
+    {
+      std::unique_lock lock(_mutex);
+      if (!isRejected(lock)) {
+        _onRejectedListeners.push_back(std::move(onRejected));
+        return;
+      }
+      error = std::get<std::exception_ptr>(_state);
     }
+    onRejected(error);
```
mrousavy: I don't like this diff. It reads much harder than the part before - before it was lock, check if it is resolved, if yes call callback now, if not add listener.
Fast path first, longer path fallback. Easy to understand. No early return, no messy RAII blocks, no default path always being hit except for early returns, no nothing.

packages/react-native-nitro-modules/cpp/core/Promise.hpp
```diff
@@ -145,49 +162,53 @@ class Promise final {
...
+    std::exception_ptr error;
+    {
+      std::unique_lock lock(_mutex);
+      if (!isRejected(lock)) {
+        _onRejectedListeners.push_back(onRejected);
+        return;
+      }
+      error = std::get<std::exception_ptr>(_state);
     }
+    onRejected(error);
```
mrousavy: I don't like this diff. It reads much harder than the part before - before it was lock, check if it is resolved, if yes call callback now, if not add listener.
Fast path first, longer path fallback. Easy to understand. No early return, no messy RAII blocks, no default path always being hit except for early returns, no nothing.

packages/react-native-nitro-modules/cpp/core/Promise.hpp
```diff
@@ -231,25 +254,50 @@ class Promise final {
...
   }
   /**
    * Gets whether this Promise has not yet been resolved nor rejected.
    */
   [[nodiscard]]
   inline bool isPending() const noexcept {
-    return std::holds_alternative<std::monostate>(_state);
+    std::unique_lock lock(_mutex);
+    return isPending(lock);
   }
```
mrousavy, line 279: Do we even need those methods? I mean publicly facing you would probably only use `onResolvedListener` anyways, no?

mrousavy (changes requested):
Hey - thanks for the PR, this is good stuff - I have a few minor nitpicks around the code, please take a look.

packages/react-native-nitro-modules/cpp/core/Promise.hpp
mfazekas: Yes — `JSIConverter+Promise.hpp` uses all three (`isPending`, `isResolved`, `isRejected`) to decide whether to call the JS resolve/reject synchronously or to add listeners. Removed the helpers and restructured the listeners as suggested in the other comments.

packages/react-native-nitro-modules/cpp/core/Promise.hpp
mfazekas: Calling a callback in a lock is playing with fire.  This is what this change is about. 

Not sure if we can keep that and the original structure. We can do this, if you prefer:

```cpp
 void addOnRejectedListener(OnRejectedFunc&& onRejected) {
    std::unique_lock lock(_mutex);
    if (std::holds_alternative<std::exception_ptr>(_state)) {
      std::exception_ptr error = std::get<std::exception_ptr>(_state);
      lock.unlock();          // unlocks here
      onRejected(error);      // callback runs without lock
    } else {
      _onRejectedListeners.push_back(std::move(onRejected));
    }
  }
```

packages/react-native-nitro-modules/cpp/core/Promise.hpp
mrousavy: Good point, okay then we should definitely keep it.

Alternatively we could make the JSIConverter conformance part of Promise.hpp, and make it `friend` methods to avoid leaking into public API - that would also be a minor speedup as we would only need to lock once then, not 3 times in JSIConverter. But that's for future optimization

packages/react-native-nitro-modules/cpp/core/Promise.cpp
```diff
@@ -52,62 +52,82 @@ std::shared_ptr<Promise<void>> Promise<void>::rejected(const std::exception_ptr&
...
   assertPromiseState(*this, PromiseTask::WANTS_TO_REJECT);
 #endif
-  _error = exception;
-  for (const auto& onRejected : _onRejectedListeners) {
+  std::vector<OnRejectedFunc> listeners;
+  {
+    std::unique_lock lock(_mutex);
+    _error = exception;
+    listeners = std::move(_onRejectedListeners);
+    clearAllListeners(lock);
```
mrousavy: Any reason why we have to do that in here anyways? Why not do it later without the lock? 
`_error` (or `_result` / `_isResolved`) is assigned under a lock anyways, any future calls to `addOnResolvedListener` will no longer add listeners to our list anyways, so we can clear the list and will never have any more items in it, right?

Point being less work performed under a lock. Destruction can take some time too, especially if it's shared_ptrs that do locking or atomics, or BorrowingReference.

mrousavy: Basically moving `clearAllListeners` down to where `didFinish` was before.

Just an idea, not necessarily a requirement. Maybe you have good reason to not do that

packages/react-native-nitro-modules/cpp/core/Promise.hpp
```diff
@@ -98,26 +98,38 @@ class Promise final {
...
   void resolve(TResult&& result) {
-    std::unique_lock lock(_mutex);
 #ifdef NITRO_DEBUG
     assertPromiseState(*this, PromiseTask::WANTS_TO_RESOLVE);
 #endif
-    _state = std::move(result);
-    for (const auto& onResolved : _onResolvedListeners) {
-      onResolved(std::get<TResult>(_state));
+    std::vector<OnResolvedFunc> listeners;
+    const TResult* resolvedValue;
```
mrousavy, line 106: Any reason why this is a pointer? I mean it will point to `this`, even later when the lock is unlocked - right?

Can we not make this `TResult` and move the `_state` into it?

mrousavy: Ah sorry nevermind that makes no sense, then we don't have `_state` anymore if we move it out.

packages/react-native-nitro-modules/cpp/core/Promise.hpp
```diff
@@ -98,26 +98,38 @@ class Promise final {
...
+    {
+      std::unique_lock lock(_mutex);
+      _state = result;
+      resolvedValue = &std::get<TResult>(_state);
+      listeners = std::move(_onResolvedListeners);
+      clearAllListeners(lock);
+    }
+    for (const auto& onResolved : listeners) {
+      onResolved(*resolvedValue);
```
mrousavy:
```suggestion
    std::vector<OnResolvedFunc> listeners;
    {
      std::unique_lock lock(_mutex);
      _state = result;
      listeners = std::move(_onResolvedListeners);
      clearAllListeners(lock);
    }
    for (const auto& onResolved : listeners) {
      onResolved(result);
```

packages/react-native-nitro-modules/cpp/core/Promise.hpp
mfazekas: It's used in one place, so I might make sense to take a lock once and get both the state and resolved result or error in once call - one lock for convert: https://github.com/mfazekas/nitro/blob/eb139ff256b298d45c56413fff79e45be20f437d/packages/react-native-nitro-modules/cpp/jsi/JSIConverter+Promise.hpp#L56-L93

Should I rewrite?!

packages/react-native-nitro-modules/cpp/core/Promise.cpp
mfazekas: So I adopter this pattern, in all 5 place:

```cpp
void Promise<void>::reject(const std::exception_ptr& exception) {
  ...
  std::vector<OnRejectedFunc> listeners;
  std::vector<OnResolvedFunc> dropped;
  {
    std::unique_lock lock(_mutex);
    ...
    listeners = std::move(_onRejectedListeners);
    dropped = std::move(_onResolvedListeners);
  }
  for (const auto& onRejected : listeners) {
    onRejected(exception);
  }
}
```

packages/react-native-nitro-modules/cpp/core/Promise.hpp
mrousavy: Not in this PR - maybe I will do that in a future PR.

packages/react-native-nitro-modules/cpp/core/Promise.hpp
mrousavy: Ah yes that makes sense we dont want the callback to run under the lock...

mrousavy:
letsgo! thanks

PR #1411, opened 2026-06-23
chore: Enable native crash detection in Harness CI

mrousavy:


mrousavy:
its already enabled by default

PR #1413, opened 2026-06-24
fix(threading): read _tasks.empty() under _queueMutex in ThreadPool::run()

mfazekas:
`ThreadPool::run()` checked `_tasks.empty()` before acquiring `_queueMutex`, while worker threads call `_tasks.pop()` under that same mutex. `std::queue` is backed by `std::deque` — even `empty()` reads internal size state that `pop_front()` writes non-atomically, so this is a real data race caught by ThreadSanitizer.

Fix: move the `_tasks.empty()` check inside the existing lock scope. `addThread()` is then called after releasing the lock since it also takes `_queueMutex`.

**TSan report** (from CI on [mfazekas/nitro#2](https://github.com/mfazekas/nitro/pull/2), after the `Promise<T>` race was fixed):

```
WARNING: ThreadSanitizer: data race (pid=30664)
  Read of size 8 at 0x0001033cb400 by thread T8:
    #0 std::deque<function<void()>>::size() const  deque:756
    #1 std::deque<function<void()>>::empty() const  deque:767
    #2 std::queue<function<void()>>::empty() const  queue:378
    #3 margelo::nitro::ThreadPool::run(function<void()>&&)  ThreadPool.cpp:62
    #4 margelo::nitro::Promise<double>::async(function<double()>&&)  Promise.hpp:57
    #5 margelo::nitro::test::HybridTestObjectCpp::promiseReturnsInstantlyAsync()  HybridTestObjectCpp.cpp:559

  Previous write of size 8 at 0x0001033cb400 by thread T15 (mutexes: write M0):
    #0 std::deque<function<void()>>::pop_front()  deque:2302
    #1 std::queue<function<void()>>::pop()  queue:415
    #2 margelo::nitro::ThreadPool::addThread()::lambda  ThreadPool.cpp:51

  Location is global 'margelo::nitro::ThreadPool::shared()::shared'
```

Fix verified clean on [mfazekas/nitro#3](https://github.com/mfazekas/nitro/pull/3) — TSan job passes with no reports.

### DIFF
--- a/packages/react-native-nitro-modules/cpp/threading/ThreadPool.cpp
+++ b/packages/react-native-nitro-modules/cpp/threading/ThreadPool.cpp
@@ -16,13 +16,13 @@ ThreadPool::ThreadPool(const char* name, size_t initialThreadsCount, size_t maxT
   Logger::log(LogLevel::Info, TAG, "Creating ThreadPool \"%s\" with %i initial threads (max: %i)...", name, initialThreadsCount,
               maxThreadsCount);
 
+  std::unique_lock<std::mutex> lock(_queueMutex);
   for (size_t i = 0; i < initialThreadsCount; i++) {
-    addThread();
+    addThread(lock);
   }
 }
 
-void ThreadPool::addThread() {
-  std::unique_lock<std::mutex> lock(_queueMutex);
+void ThreadPool::addThread(std::unique_lock<std::mutex>& queueMutexLock) {
   if (!_isAlive) {
     return;
   }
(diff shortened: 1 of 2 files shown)


mrousavy (changes requested):
Thanks for the PR, this is great stuff! I wanted to add TSan to the CI to iOS + Android runs to catch such issues in the future in the CI.

One minor nitpick for this PR; RAII has both upsides and downsides. Upside is it's predictable and non leaking, downside is that it sometimes causes the code to look ugly - like here, `bool shouldAddThread` leaks outside of the current scope and will be used later, kind of an ugly pattern.

How about we just call `addThread()` just above `_tasks.push(...)` under our `lock`, and inside `addThread()` we remove the `std::unique_lock`?

That would also speed up `ThreadPool` initialization since we no longer need to lock a few times to spawn the initial Threads - right?

mrousavy:
thanks lgtm

efstathiosntonas:
hey guys, sorry for hijacking this pr but the latest version should be 0.35.10 (as per npmjs) not 0.35.11 on https://github.com/mrousavy/nitro/releases/tag/v0.35.11

mrousavy:
oh, whoops thanks

PR #1422, opened 2026-06-26
Add dedicated catch paths for jsi and std::runtime exceptions.

### LINKED ISSUE #1421: `jsi::JSError` thrown inside Nitro hybrid methods is caught in the catch(...) all block leading to loss of original message.
### What happened?

**File:** `cpp/core/HybridFunction.hpp` (~line 117)
**Upstream repo:** https://github.com/mrousavy/nitro

## Problem

When a `jsi::JSError` is thrown inside a Nitro hybrid method (e.g., during jsi value conversion), the catch block in `HybridFunction.hpp` catches it in the catch all block. It then re-wraps the error with method name information, replacing the original error message and on js side we receive exception name as `N8facebook3jsi7JSErrorE`

A similar issue applies to `std::runtime_error`: it gets caught by the generic
handler and unnecessarily decorated with method-name context, when the raw message
is more useful. We throw std::runtime_error in our cpp implemented methods.

## Fix

Adding dedicated catch blocks for jsi::JSError and std::runtime solves the issue and we get proper messages on the JS side.

### Reproduceable Code

```tsx
export interface Logger extends HybridObject<{ ios: 'c++'; android: 'c++' }> {
  // Initialization
  configure(config: LoggerConfiguration): void;
  shutdown(): void;

  // Logging methods
  trace(tag: string, message: string): void;
  debug(tag: string, message: string): void;
  info(tag: string, message: string): void;
  error(tag: string, message: string): void;

  // Structured logging
  log(level: TLogLevel, tag: string, message: string): void;

  // Context management
  setGlobalContext(context: LoggerContext): void;
  clearGlobalContext(): void;
}


now call any of the methods from JS side,
NitroModules.createHybridObject<T>('T').info('asdf', 3); // or pass undefined / null / boolean (anything non-string)

We get this exception on JS console
jsi::JSError("`Logger.info(...)` threw an unknown
                                                N8facebook3jsi7JSErrorE error.")
```

### Relevant log output

```shell
jsi::JSError("`Logger.info(...)` threw an unknown
                                                N8facebook3jsi7JSErrorE error.")
```

### Device

All devices.

### Nitro Modules Version

0.32.1

### Nitrogen Version

0.33.2

### Additional information

- [ ] I am using Expo
- [x] I am using nitrogen

07abhishekrock:
Fixes : https://github.com/mrousavy/nitro/issues/1421

### DIFF
--- a/packages/react-native-nitro-modules/cpp/core/HybridFunction.hpp
+++ b/packages/react-native-nitro-modules/cpp/core/HybridFunction.hpp
@@ -117,6 +117,12 @@ public:
         // 3. Actually call the method with JSI values as arguments and return a JSI value again.
         //    Internally, this method converts the JSI values to C++ values using `JSIConverter<T>`.
         return callMethod(hybridInstance.get(), method, runtime, args, count, std::index_sequence_for<Args...>{});
+      } catch(const std::runtime_error& e) {
+          std::string funcName = getHybridFuncFullName<THybrid>(kind, name, hybridInstance.get());
+          throw jsi::JSError(runtime, funcName + ": " + e.what());
+      } catch(const jsi::JSError& e) {
+          std::string funcName = getHybridFuncFullName<THybrid>(kind, name, hybridInstance.get());
+          throw jsi::JSError(runtime, funcName + " threw an error: " + e.getMessage());
       } catch (const std::exception& exception) {
         // Some exception was thrown - add method name information and re-throw as `JSError`.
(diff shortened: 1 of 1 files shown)


mrousavy:
std runtime error is a subclass of std exception, so the std exception catch block _should_ capture std runtime error too, unless you accidentally link libc++ twice or dont enable -fexceptions or -frtti. This was an issue in an old rn version and we fixed it upstream.

Thank you for your PR and time spent on this, but unfortunately this is not a proper fix, so closing

PR #1423, opened 2026-06-26
fix: raw Props arguments issue

riteshshukla04:
Raw props accept dont accept 3 arguments now . This is causing Viison Camera and nitro image to fail in nightly . See https://github.com/react/react-native/commit/67381e157faadd9f18db1ff8710d8d46c3c1dae5


mrousavy:
Hey thanks for this PR! Unfortunately it doens't build. Also, I think it's unnecessarily using templates, which makes code harder to read, longer to compile, and isn't a TU but instead a header. I created https://github.com/mrousavy/nitro/pull/1425 which I will favor for this fix

PR #1424, opened 2026-06-26
fix: version-gate generated RawProps::at() for react-native 0.87

cortinico:
react-native 0.87 removed the (name, prefix, suffix) overload of
react::RawProps::at() in favor of a single-argument at(name):

- https://github.com/react/react-native/pull/55763

This breaks nitro codegen output on newer react-native while older versions
(<= 0.86) still require the 3-argument form.

Update the CppHybridViewComponent codegen template to emit a
NITRO_RAWPROPS_AT macro guarded by REACT_NATIVE_VERSION_* (from
<cxxreact/ReactNativeVersion.h>):

  * RN > 0.86 (incl. nightly 1000.x): rawProps.at(name)
  * RN <= 0.86:                       rawProps.at(name, nullptr, nullptr)

Regenerate the committed HybridTestView / HybridRecyclableTestView
component files so the generated output stays in sync with the template.

### DIFF
--- a/packages/nitrogen/src/views/CppHybridViewComponent.ts
+++ b/packages/nitrogen/src/views/CppHybridViewComponent.ts
@@ -209,7 +209,7 @@ namespace ${namespace} {
       `
 ${name}([&]() -> CachedProp<${type}> {
   try {
-    const react::RawValue* rawValue = rawProps.at("${prop.name}", nullptr, nullptr);
+    const react::RawValue* rawValue = NITRO_RAWPROPS_AT(rawProps, "${prop.name}");
     if (rawValue == nullptr) return sourceProps.${name};
     const auto& [runtime, value] = (std::pair<jsi::Runtime*, jsi::Value>)*rawValue;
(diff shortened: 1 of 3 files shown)


mrousavy:
Hey Nico thanks for this PR! Really appreciate it ❤️ 
While it seems to fix the compat, I'm not sure if I like C/C++ macros (`#define` + `#undef`) in this case. 

There is no super elegant fix for this, but I just created https://github.com/mrousavy/nitro/pull/1425 which ships this in a `.cpp` file, allowing us to potentially change params if this API surface is expected to change again in the future.

Btw., this begs the question - why don't we keep a `RawProps::at(char*, char*, char*)` overload in RN core?

cortinico:
> Btw., this begs the question - why don't we keep a `RawProps::at(char*, char*, char*)` overload in RN core?

Yeah we're re-adding it but we intend to remove it in a future version

mrousavy:
Fixed via https://github.com/mrousavy/nitro/pull/1491

PR #1425, opened 2026-06-29
fix: Add support for new (+ old) `RawProps::at` overload for react-native 0.86

mrousavy:


mrousavy:
Fixed via https://github.com/mrousavy/nitro/pull/1491

PR #1426, opened 2026-06-29
feat: Support variants of union-style enums

mrousavy:
Adds support for creating Variants of union-style enums as well.

Example:

```ts
type CameraPosition = "front" | "back" | "external"
interface CameraDevice extends HybridObject<...> { ... }

interface Config {
  input: CameraDevice | CameraPosition
}
```

Previously, this would've flattened `input` to `CameraDevice | 'front' | 'back' | 'external'` and therefore threw:

```diff
- ❌  Failed to generate spec for CameraSession! Error: String literal "external" cannot be represented in C++ because it is ambiguous between a string and a discriminating union enum.
```

Now this properly preserves the `CameraPosition` union type, and generates the variant properly.


packages/nitrogen/src/syntax/createType.ts
```diff
@@ -171,7 +223,8 @@ export function addKnownType(
 export function createType(
   language: Language,
   type: TSMorphType,
-  isOptional: boolean
+  isOptional: boolean,
+  typeNode?: TypeNode
```
mrousavy, line 227: I really don't like the fact that we have to pass in a new optional parameter (`TypeNode`) here now. This seems super hacky.
But not sure if we can do anything to prevent this tbh.

PR #1432, opened 2026-07-08
fix: Fix clearing optional view props crashing with "Value is null"

### LINKED ISSUE #1184: Error when assigning `undefined` to optional view property

mfazekas:
Fixes #1184, closes #1189 (the alternative converter-level fix): instead of loosening `JSIConverter<std::optional<T>>` for *all* conversions, this maps `null` to "no value" **only for view props** - `CachedProp::fromRawValue` treats `null` like `undefined`, unless the prop's type explicitly models null (e.g. `string | null`). Method arguments and hybrid object properties stay strict about `null` vs `undefined`.

For view props that distinction doesn't exist anyway: React sends `null` both for a removed prop ([ReactNativeAttributePayload.js#L390-L394](https://github.com/facebook/react-native/blob/v0.85.3/packages/react-native/Libraries/ReactNative/ReactFabricPublicInstance/ReactNativeAttributePayload.js#L390-L394)) and for an explicit `undefined` ([#L270-L277](https://github.com/facebook/react-native/blob/v0.85.3/packages/react-native/Libraries/ReactNative/ReactFabricPublicInstance/ReactNativeAttributePayload.js#L270-L277)), so the native side cannot tell them apart - it just crashed the commit with `Exception in HostFunction: <View>.<prop>: Value is null, expected a String`.

Also covers optional callback props (not fixed by #1189): those crashed earlier, in the generated `{ f: ... }` unwrap (`value.asObject()`), before any converter ran.

Notes:

- For props that model null (`foo?: string | null`), clearing the prop delivers `null`, not `undefined` - React's payload makes the two indistinguishable, and null-modeling types take `null` at face value. Pinned by a test.
- Required props are untouched: a `null` payload still throws, with the original "Value is null" message.
- The callback-prop part lives in generated code, so libraries pick it up by re-running nitrogen; the `CachedProp` part ships with `react-native-nitro-modules` directly.

Tests: optional props on the existing `TestView` + on-device harness tests in `example/__tests__/views.harness.tsx` (clearing a prop needs React's commit path, so this can't live in `getTests.ts`). They fail on `main` (#1433) and pass here, verified on iOS and Android.

## Checklist

- [x] Targets `main`, narrow scope.
- [x] Test pins the bug: fails on `main` (#1433), passes with the fix.
- [x] `bun specs` run, generated files committed.
- [x] `bun lint-all` passes locally.
- [x] No existing specs or test cases removed, no unrelated refactors.
- [x] Conventional Commits.

### DIFF
--- a/example/__tests__/views.harness.tsx
+++ b/example/__tests__/views.harness.tsx
@@ -1,0 +1,185 @@
+import * as React from 'react'
+import {
+  describe,
+  it,
+  expect,
+  render,
+  cleanup,
+  waitFor,
+  afterEach,
+} from 'react-native-harness'
+import { callback } from 'react-native-nitro-modules'
+import { TestView, type TestViewRef } from 'react-native-nitro-test'
+
+type TestViewElementProps = React.ComponentProps<typeof TestView>
+type OptionalTestViewProps = Partial<
+  Pick<
+    TestViewElementProps,
+    'optionalString' | 'nullableString' | 'optionalCallback'
+  >
+>
+
+// Prop-parsing errors are thrown during React's commit and are only
+// observable through an error boundary - the test-runner's own boundary
+// would swallow them and remount, masking the failure.
+class CatchRenderErrors extends React.Component<
+  { children: React.ReactNode; onError: (error: unknown) => void },
+  { hasError: boolean }
+> {
+  state = { hasError: false }
+  static getDerivedStateFromError() {
+    return { hasError: true }
+  }
+  componentDidCatch(error: unknown) {
+    this.props.onError(error)
+  }
+  render() {
+    return this.state.hasError ? null : this.props.children
+  }
+}
+
+// React encodes "prop was removed" as `null` (not `undefined`) in the Fabric
+// update payload, which used to crash Nitro's prop parsing:
+// https://github.com/mrousavy/nitro/issues/1184
+describe('TestView optional props', () => {
+  afterEach(() => {
+    cleanup()
+  })
+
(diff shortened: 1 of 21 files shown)


mrousavy:
Overall this PR is quite good and fixes a valid bug - thank you!

I will investigate a bit further to see if we can get the prop converter fully into Nitro core so that it's not a burden of the generated code to properly convert props. That includes the `f` unwrapping of functions, seems like there's more stuff we need to do to support Views than just unwrap props :)

PR #1506, opened 2026-08-19
fix: Make View prop snapshots truly immutable via diffing

mrousavy:
Before, we did a lot of messy voodoo magic like `const_cast` to make React `Props` mutable (even though they are immutable by design), and then tracked a `isDirty` flag on them which was cleared after we set a prop in Swift/Kotlin.

This had a few issues;

1. `Props` really should be immutable (`const Props`) by React's design. Making it mutable could cause issues with React internals, or breaks in future versions
2. `isDirty` can _theoretically_ go out of sync (potentially with concurrent rendering?). In practice, that would probably never happen today because React creates new `Props` everytime.
3. I _think_ `isDirty` didn't even work before on Android (because we couldn't set it back on `Props` via `State`), **so this is now a performance improvement for Nitro Views on Android too**.

### DIFF
--- a/packages/nitrogen/src/views/swift/SwiftHybridViewManager.ts
+++ b/packages/nitrogen/src/views/swift/SwiftHybridViewManager.ts
@@ -26,8 +26,12 @@ export function createSwiftHybridViewManager(
   const { HybridTSpec, HybridTSpecSwift, HybridTSpecCxx } = getHybridObjectName(
     spec.name
   )
-  const { component, descriptorClassName, propsClassName } =
-    getViewComponentNames(spec)
+  const {
+    component,
+    descriptorClassName,
+    propsClassName,
+    shadowNodeClassName,
+  } = getViewComponentNames(spec)
   const implementation = spec.config.getIosAutolinkedImplementation(spec.name)
   if (implementation?.language !== 'swift') {
     throw new Error(
@@ -45,12 +49,13 @@ export function createSwiftHybridViewManager(
     )
     return `
 // ${p.jsSignature}
-if (newViewProps.${name}.isDirty) {
+if (oldViewProps == nullptr || !newViewProps.${name}.hasSameValue(oldViewProps->${name})) {
   swiftPart.${setter}(${indent(parse, '  ')});
-  newViewProps.${name}.isDirty = false;
 }
 `.trim()
   })
+  const hasTransactionPropChanges =
+    'oldViewProps == nullptr || !newViewProps.hasSameProps(*oldViewProps)'
 
   const mmFile = `
 ${createFileMetadataString(`${component}.mm`)}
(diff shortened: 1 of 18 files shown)


packages/nitrogen/src/views/swift/SwiftHybridViewManager.ts
```diff
@@ -45,12 +49,13 @@ export function createSwiftHybridViewManager(
...
   swiftPart.${setter}(${indent(parse, '  ')});
-  newViewProps.${name}.isDirty = false;
 }
 `.trim()
   })
+  const hasTransactionPropChanges =
+    'oldViewProps == nullptr || !newViewProps.hasSameProps(*oldViewProps)'
 
   const mmFile = `
 ${createFileMetadataString(`${component}.mm`)}
```
mrousavy, line 58:
```suggestion
})

const mmFile = `
${createFileMetadataString(`${component}.mm`)}
```

packages/nitrogen/src/views/swift/SwiftHybridViewManager.ts
```diff
@@ -136,28 +142,30 @@ using namespace ${namespace}::views;
...
-  const ${propsClassName}& newViewPropsConst = *std::static_pointer_cast<${propsClassName} const>(props);
-  ${propsClassName}& newViewProps = const_cast<${propsClassName}&>(newViewPropsConst);
+  const auto& newViewProps = *std::static_pointer_cast<const ${propsClassName}>(props);
+  const auto* oldViewProps = static_cast<const ${propsClassName}*>(oldProps.get());
   ${swiftNamespace}::${HybridTSpecCxx}& swiftPart = _hybridView->getSwiftPart();
 
-  // 2. Update each prop individually
-  swiftPart.beforeUpdate();
+  // 2. Update only props that differ from the previous Props snapshot.
+  const bool hasTransactionPropChanges = ${hasTransactionPropChanges};
```
mrousavy:
```suggestion
  const bool hasTransactionPropChanges = oldViewProps == nullptr || !newViewProps.hasSameProps(*oldViewProps);
```

packages/nitrogen/src/views/swift/SwiftHybridViewManager.ts
```diff
@@ -45,12 +49,19 @@ export function createSwiftHybridViewManager(
     )
     return `
 // ${p.jsSignature}
-if (newViewProps.${name}.isDirty) {
+if (oldViewProps == nullptr || !newViewProps.${name}.hasSameValue(oldViewProps->${name})) {
```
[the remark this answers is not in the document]
mrousavy: @chatgpt-codex-connector Fixed via https://github.com/mrousavy/nitro/pull/1510 (I think)

PR #1531, opened 2026-08-26
fix(nitrogen): qualify iOS bridge headers

### LINKED ISSUE #1293: Nitrogen iOS bridge headers use quoted includes for shared C++ types — collides when two Nitro modules share a header filename
### Summary

Nitrogen-generated iOS Cxx bridge headers use quoted `#include "Foo.hpp"` for shared C++ types from the same Nitro module. When two Nitro modules in the same app generate a header with the same filename, the static-framework header layout can resolve the include to the wrong file or fail with `'Foo.hpp' file not found`.

### Concrete trigger

Hit while building an Expo SDK 55 / RN 0.83.4 app for TestFlight via EAS:

- `react-native-vision-camera@5.0.1` generates `nitrogen/generated/shared/c++/Orientation.hpp`
- `react-native-unistyles@3.2.3` also generates `nitrogen/generated/shared/c++/Orientation.hpp`

Two unrelated `enum class Orientation` types in different namespaces. Each library's podspec adds its own `nitrogen/generated/shared/c++` to `HEADER_SEARCH_PATHS`, and CocoaPods copies each file via `CpHeader` into the framework Headers folder:

```
Unistyles.framework/Headers/Orientation.hpp
VisionCamera.framework/Headers/Orientation.hpp
```

VisionCamera's Nitrogen-generated bridge files contain quoted includes like:

```cpp
#include "Orientation.hpp"
```

These fail compilation with `useFrameworks: 'static'`:

```
node_modules/react-native-vision-camera/nitrogen/generated/ios/VisionCamera-Swift-Cxx-Umbrella.hpp:302:10: fatal error: 'Orientation.hpp' file not found
node_modules/react-native-vision-camera/nitrogen/generated/ios/VisionCamera-Swift-Cxx-Bridge.hpp:304:10: fatal error: 'Orientation.hpp' file not found
node_modules/react-native-vision-camera/nitrogen/generated/ios/c++/HybridCameraOutputSpecSwift.hpp:21:10: fatal error: 'Orientation.hpp' file not found
```

— and four more sibling `Hybrid*Swift.hpp` files. Seven generated headers in total in vision-camera 5.0.1 alone.

Filed downstream at mrousavy/react-native-vision-camera#3747 with the patch.

### Why this is a codegen problem, not a vision-camera or unistyles problem

The generic name `Orientation` is reasonable for both libraries — one is screen orientation, the other is camera/frame orientation. Nothing prevents two future Nitro libs from naming a shared type the same: `State`, `Event`, `Region`, `Color`, `Mode`, `Size`, etc.

The codegen has no namespacing on filenames. It relies on quoted includes that resolve via `HEADER_SEARCH_PATHS`. As soon as both modules' header dirs end up on the search path (which is the default Pod layout), the include is structurally ambiguous. Whether it happens to compile depends on Pod resolution order and the toolchain — Xcode 26 + Hermes-from-source surfaces it deterministically; older configurations sometimes hide it.

The same class of breakage has been filed elsewhere with different surface symptoms but the same root cause:

- patlux/react-native-bluetooth-state-manager#113 — `'NitroModules/AnyMapHolder.hpp' file not found` after nitro 0.35. Workaround: hand-edit the generated umbrella to use `<NitroModules/AnyMapUtils.hpp>` (framework-form include).
- NitrogenZLab/react-native-multiple-image-picker#260 — same `AnyMapHolder.hpp` symptom on nitro 0.35.2.

Both are quoted-include resolution failures from generated umbrella headers, just with a different cause (renamed nitro internal header). Same shape — same fix shape.

### Suggested fix

For the iOS bridge files Nitrogen emits, prefer framework-form includes for any header that lives in another `nitrogen/generated/shared/c++/` of a Nitro module — including the module's own:

```diff
-#include "Orientation.hpp"
+#include <VisionCamera/Orientation.hpp>
```

Two reasons to prefer framework-form:

1. **Disambiguation.** `<ModuleName/Header.hpp>` resolves through CocoaPods modulemaps and cannot collide with another Pod's identically-named header.
2. **Stability across Pod-resolution order.** Quoted includes depend on `HEADER_SEARCH_PATHS` ordering, which CocoaPods is allowed to vary. Static-framework + Hermes-from-source + Xcode 26 happens to surface this; older configurations hid it.

The `<NitroModules/ArrayBufferHolder.hpp>` and `<NitroImage/HybridImageSpec.hpp>` lines already in those same generated files prove the pattern works under the same build. The module name is known to nitrogen at codegen time (it's in `nitro.json`), so the substitution is mechanical and entirely backward-compatible.

### Reproduction

```bash
npx create-expo-app repro
cd repro
npm i react-native-vision-camera@5.0.1 react-native-unistyles@3.2.3 \
      react-native-nitro-modules@0.35.4 react-native-nitro-image@0.13.1 \
      expo-build-properties react-native-edge-to-edge
```

`app.config.js`:

```js
plugins: [
  ['expo-build-properties', {
    ios: { useFrameworks: 'static' },
  }],
],
```

`eas build -p ios --profile production` (Xcode 26).

### Patch I'm running locally as a workaround

```diff
# patches/react-native-vision-camera+5.0.1.patch
# 7 hunks, all the same shape:
-#include "Orientation.hpp"
+#include <VisionCamera/Orientation.hpp>
```

Files touched:

- `nitrogen/generated/ios/VisionCamera-Swift-Cxx-Umbrella.hpp`
- `nitrogen/generated/ios/VisionCamera-Swift-Cxx-Bridge.hpp`
- `nitrogen/generated/ios/c++/HybridCameraOutputSpecSwift.hpp`
- `nitrogen/generated/ios/c++/HybridDepthSpecSwift.hpp`
- `nitrogen/generated/ios/c++/HybridFrameSpecSwift.hpp`
- `nitrogen/generated/ios/c++/HybridOrientationManagerSpecSwift.hpp`
- `nitrogen/generated/ios/c++/HybridPhotoSpecSwift.hpp`

Build archives and uploads to TestFlight cleanly with the patch.

Happy to open a PR against the codegen if there's interest in the framework-form approach — the change is small (substitution rule in the iOS bridge templates) and would protect against any future filename collision.

### Environment

- nitro: 0.35.4 (`react-native-nitro-modules`)
- vision-camera: 5.0.1
- unistyles: 3.2.3
- React Native: 0.83.4
- Expo: SDK 55 (auto-assigned Xcode 26 by EAS Build)
- iOS deployment target: 18

### Related

- mrousavy/react-native-vision-camera#3747 (downstream report with the patch)
- patlux/react-native-bluetooth-state-manager#113 (same root cause, different file)
- NitrogenZLab/react-native-multiple-image-picker#260 (same root cause, different file)

huytdps13400:
## Summary

- emit module-qualified framework includes for module-local shared C++ headers used by generated iOS bridges
- preserve existing system/external imports and quoted private Swift-bridge imports
- keep the previous local-before-system ordering so this change does not alter transitive include behavior
- regenerate the checked-in NitroTest and NitroTestExternal iOS fixtures

## Root cause

Nitrogen emitted shared types such as `Orientation.hpp` as quoted includes. With static frameworks, identically named public headers from multiple Nitro modules can be copied into separate framework header namespaces, while an unqualified include remains ambiguous and may resolve to the wrong module.

The iOS bridge headers now emit `<IosModule/Header.hpp>`, so CocoaPods resolves each type through the owning module.

## Regression coverage

The existing NitroTest specs already exercise local structs, enums, HybridObject specs, and a second external Nitro module. Their checked-in generated output now contains 82 module-qualified local includes. `bun specs` is the golden regression: reverting the generator changes these fixtures and fails the CI regeneration check.

## Verification

- `bun specs`
- `bun run build`
- `bun typecheck`
- all five JS/TS `lint-ci` package checks
- `bun nitro-test test --runInBand`
- Xcode 26.6 iOS example build with `USE_FRAMEWORKS=static`: **BUILD SUCCEEDED**
- Xcode 26.6 iOS example build without frameworks: **BUILD SUCCEEDED**
- generated-output sweep: no iOS header uses a quoted include for a module-local `generated/shared/c++` header
- `git diff --check`

Fixes #1293


eduardoborges:
Two more collisions from an RN 0.87 app using `use_frameworks! :linkage => :static`:

- `ResizeMode.hpp`, between react-native-video 7.0.0-beta.11 and react-native-nitro-image (TheWidlarzGroup/react-native-video#4912)
- `ListenerSubscription.hpp`, between react-native-video 7.0.0-beta.11 and react-native-vision-camera 5.2.3. This one only broke on a clean Xcode Cloud build. Locally it compiled.

We patch both libs to include those headers through the framework path. To see if this PR would let us drop the patches, I ran nitrogen from this branch and from its base (c164d24) against both packages:

| Library | Header | Base | This PR |
|---|---|---|---|
| react-native-video 7.0.0-beta.11 | `ResizeMode.hpp` (2 files) | `"ResizeMode.hpp"` | `<ReactNativeVideo/ResizeMode.hpp>` |
| react-native-vision-camera 5.2.3 | `ListenerSubscription.hpp` (5 files) | `"ListenerSubscription.hpp"` | `<VisionCamera/ListenerSubscription.hpp>` |

All 7 includes we patch by hand come out qualified, and the diff between base and branch only changes `#include` lines. I haven't rebuilt the app with the regenerated code, so this only checks the generated output.

PR #1546, opened 2026-08-30
fix: Validate native ArrayBuffer sizes

### LINKED ISSUE #1545: createNativeArrayBuffer accepts invalid allocation sizes
## Description

`createNativeArrayBuffer(double)` casts its JavaScript number directly to `size_t`. Negative, non-finite, and values beyond the platform size limit can therefore become huge allocations or enter an out-of-range floating-to-integer conversion.

## Reproduction

Call `NitroModules.createNativeArrayBuffer(-1)`, `Infinity`, or `Number.MAX_VALUE`. Current native code reaches the unchecked allocation path.

## Expected behavior

Invalid sizes should reject deterministically before conversion or allocation.

OskarEichler:
## Summary

- Reject non-finite, negative, and platform-out-of-range native ArrayBuffer sizes before conversion.
- Add native harness coverage for negative, infinite, and oversized JavaScript inputs.
- Rebase onto current main while preserving the newly added external-memory-pressure harness coverage.

## Breaking changes

Invalid sizes now reject deterministically instead of reaching an unsafe conversion/allocation. Valid sizes are unchanged.

## Verification

- Current-main root build, package/example typechecks, package lint, and `git diff --check` passed.
- Original focused branch also passed C++ format, a full iOS Simulator native build, and focused iOS Harness: 8 tests across both native test matrices; 520 unrelated tests skipped. The rebased source fix is identical; only the adjacent upstream harness test required conflict resolution.

## Preview

Not applicable; this is a nonvisual native validation fix.

Fixes #1545

### DIFF
--- a/packages/react-native-nitro-modules/cpp/entrypoint/HybridNitroModulesProxy.cpp
+++ b/packages/react-native-nitro-modules/cpp/entrypoint/HybridNitroModulesProxy.cpp
@@ -9,6 +9,8 @@
 #include "HybridObjectRegistry.hpp"
 #include "JSIConverter.hpp"
 #include "NitroDefines.hpp"
+#include <cmath>
+#include <limits>
 
 namespace margelo::nitro {
 
@@ -80,6 +82,10 @@ std::shared_ptr<HybridObject> HybridNitroModulesProxy::updateMemorySize(const st
 }
 
 std::shared_ptr<ArrayBuffer> HybridNitroModulesProxy::createNativeArrayBuffer(double size) {
+  if (!std::isfinite(size) || size < 0 || static_cast<long double>(size) > static_cast<long double>(std::numeric_limits<size_t>::max()))
+      [[unlikely]] {
+    throw std::invalid_argument("ArrayBuffer size must be a finite, non-negative number within the platform size limit!");
+  }
   return ArrayBuffer::allocate(static_cast<size_t>(size));
 }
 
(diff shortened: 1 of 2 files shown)


packages/react-native-nitro-modules/cpp/entrypoint/HybridNitroModulesProxy.cpp
```diff
@@ -80,6 +82,10 @@ std::shared_ptr<HybridObject> HybridNitroModulesProxy::updateMemorySize(const st
 }
 
 std::shared_ptr<ArrayBuffer> HybridNitroModulesProxy::createNativeArrayBuffer(double size) {
+  if (!std::isfinite(size) || size < 0 || static_cast<long double>(size) > static_cast<long double>(std::numeric_limits<size_t>::max()))
+      [[unlikely]] {
+    throw std::invalid_argument("ArrayBuffer size must be a finite, non-negative number within the platform size limit!");
+  }
   return ArrayBuffer::allocate(static_cast<size_t>(size));
```
mrousavy, line 89: Why is this only guarded in `createNativeArrayBuffer(..)` and not in `ArrayBuffer`'s constructors itself?

mrousavy (changes requested):
Overall the idea of guarding inputs is okay, but typically this is never happening in prod anyways.
Who allocates `Infinity` sized ArrayBuffers?

PR #1548, opened 2026-08-30
fix: Allow nullable Kotlin Promise results

### LINKED ISSUE #1547: Nullable Kotlin Promise results reject instead of resolving null
## Description

Kotlin `Promise<T>.then` uses `boxedResult as? T ?: throw`, which cannot distinguish a valid `null` result for nullable `T` from a failed safe cast. A `Promise<Double?>` that resolves null therefore rejects when `.then`/`.await` consumes it.

## Reproduction

Resolve `Promise<Double?>` with `null`, await it inside another Nitro promise, and return that promise to JavaScript. Current Android execution rejects with `Failed to cast Object to T!`.

## Expected behavior

Nullable Promise types should resolve null successfully through Kotlin continuation and await paths.

OskarEichler:
## Summary

- Preserve valid null values when Kotlin `Promise<T>` uses a nullable generic type.
- Add a generated cross-platform `awaitNullablePromise` spec and runtime harness regression.
- Regenerate bindings with current Nitrogen 0.37.1.

## Breaking changes

None. Nullable promise results now fulfill their declared contract instead of rejecting.

## Verification

- Frozen install/postinstall, Nitrogen regeneration, root build, and all-workspace typecheck passed.
- Full C++/Swift/Kotlin/TypeScript lint and format checks passed.
- Android Debug build passed.
- Focused Android Harness: 2 tests passed across both native implementations; 544 unrelated tests skipped.
- `git diff --check` passed.

## Preview

Not applicable; this is a nonvisual native runtime correction.

Fixes #1547

### DIFF
--- a/packages/react-native-nitro-test/nitrogen/generated/android/kotlin/com/margelo/nitro/test/HybridTestObjectSwiftKotlinSpec.kt
+++ b/packages/react-native-nitro-test/nitrogen/generated/android/kotlin/com/margelo/nitro/test/HybridTestObjectSwiftKotlinSpec.kt
@@ -34,13 +34,13 @@ abstract class HybridTestObjectSwiftKotlinSpec: HybridObject() {
   @get:DoNotStrip
   @get:Keep
   abstract val thisObject: HybridTestObjectSwiftKotlinSpec
-
+  
   @get:DoNotStrip
   @get:Keep
   @set:DoNotStrip
   @set:Keep
   abstract var optionalHybrid: HybridTestObjectSwiftKotlinSpec?
-
+  
   @get:DoNotStrip
   @get:Keep
   @set:DoNotStrip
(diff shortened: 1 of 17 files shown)


packages/react-native-nitro-test/nitrogen/generated/android/kotlin/com/margelo/nitro/test/HybridTestObjectSwiftKotlinSpec.kt
```diff
@@ -34,13 +34,13 @@ abstract class HybridTestObjectSwiftKotlinSpec: HybridObject() {
   @get:DoNotStrip
   @get:Keep
   abstract val thisObject: HybridTestObjectSwiftKotlinSpec
-  
+
```
mrousavy: Intentional?

mrousavy (approved):
overall a good change yes. Let's see if tests run. You might want to fix formatting

packages/react-native-nitro-test/nitrogen/generated/android/kotlin/com/margelo/nitro/test/HybridTestObjectSwiftKotlinSpec.kt
OskarEichler: No—those whitespace-only changes were stale generated output. I regenerated the files, restored the generator’s exact formatting, and pushed the sync.

PR #1550, opened 2026-08-30
fix: Surface Worklets initialization errors

### LINKED ISSUE #1549: Worklets integration initialization errors are silently swallowed
## Description

`installWorkletsSupport()` wraps optional module loading, Nitro proxy boxing, and custom serializer registration in one blanket catch. This is useful when `react-native-worklets` is absent, but it also silently disables an installed integration when Nitro boxing or Worklets registration fails.

## Reproduction

With the Worklets module available, make `NitroModules.box` or `registerCustomSerializable` throw during initialization. Importing Nitro completes without surfacing the actionable failure, and Worklets support remains disabled.

## Expected behavior

Only optional module absence/loading should use the no-integration fallback; initialization failures after a successful load should propagate.

OskarEichler:
## Summary

- Limit the optional Worklets fallback to module loading.
- Let Nitro boxing and custom serializer registration failures propagate.
- Add focused regressions for both previously hidden initialization errors.

## Breaking changes

None. An installed but broken integration now reports its initialization error instead of silently disabling itself.

## Verification

- Root build and all-workspace typechecks passed.
- Full C++/Swift/Kotlin/TypeScript lint and format checks passed.
- Package Jest suite and `git diff --check` passed.

## Preview

Not applicable; this is a nonvisual initialization correction.

Fixes #1549

### DIFF
--- a/packages/react-native-nitro-modules/src/__tests__/index.test.tsx
+++ b/packages/react-native-nitro-modules/src/__tests__/index.test.tsx
@@ -1,1 +1,37 @@
-it.todo('write a test')
+jest.mock('../NitroModules', () => ({
+  NitroModules: { box: jest.fn(() => ({ unbox: jest.fn() })) },
+}))
+jest.mock('react-native-worklets', () => ({
+  registerCustomSerializable: jest.fn(),
+}))
+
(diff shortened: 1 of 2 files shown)


mrousavy (approved):
overall a good change yea okay

PR #1556, opened 2026-08-30
fix: serialize Promise settlement

### LINKED ISSUE #1553: Promise settlement is only guarded in debug builds
## Description

`Promise<T>` and `Promise<void>` guard duplicate settlement through `assertPromiseState`, but that check is compiled only when `NITRO_DEBUG` is enabled. In release builds, a second resolve/reject can overwrite state, and concurrent settlers can both pass the pre-lock state check.

## Expected behavior

Settlement should be single-shot and serialized in every build configuration. A second resolve/reject should throw without changing the original result or rejection.

## Reproduction

Create an already-resolved or rejected Promise and settle it again in a non-debug build. The duplicate settlement currently succeeds because the assertion is absent.

OskarEichler:
## Summary

- enforce single-shot settlement for `Promise<T>` and `Promise<void>` in every build configuration
- validate and transition Promise state under the same mutex, closing concurrent resolve/reject races
- preserve the original result or rejection after rejected duplicate settlements
- add a generated C++ regression exercised through the example harness

## Breaking changes

None. Invalid duplicate Promise settlement now throws consistently in release builds as it already did in debug builds.

## Verification

- root build, typecheck, lint, and C++ lint passed
- Android Debug focused harness passed (1/1)
- Android Release assembled successfully
- iOS Debug Simulator full build and focused harness passed (1/1)
- `git diff --check` passed

Fixes #1553

### DIFF
--- a/example/src/getTests.ts
+++ b/example/src/getTests.ts
@@ -1554,6 +1554,17 @@ export function getTests(
           .didNotThrow()
           .equals(true)
     ),
+    ...(testObject.name === 'TestObjectCpp'
+      ? [
+          createTest('Promise settlement is single-shot', () =>
+            it(() =>
+              (testObject as TestObjectCpp).duplicatePromiseSettlementThrows()
+            )
+              .didNotThrow()
+              .equals(true)
+          ),
+        ]
+      : []),
     createTest('promiseThatResolvesVoidInstantly() works', async () =>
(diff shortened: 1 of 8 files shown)


mrousavy (changes requested):
Why? What is the benefit? We already catch this issue and we catch it debug only, meaning fast by default in release.

PR #1557, opened 2026-08-30
fix: honor Nitrogen config path

### LINKED ISSUE #1554: Nitrogen ignores the --config option
## Description

The Nitrogen CLI advertises `--config`, but the parsed path is never passed to the configuration loader. `NitroConfig.current` always reads `./nitro.json`, so invoking Nitrogen from another directory with `--config path/to/nitro.json` still fails or uses the wrong project configuration.

## Expected behavior

The CLI should load the exact file supplied through `--config`, while preserving `./nitro.json` as the default.

OskarEichler:
## Summary

- load `NitroConfig` from the CLI's parsed `--config` path before generation starts
- retain `./nitro.json` as the default path
- make custom config paths work when Nitrogen is invoked outside the module directory

## Breaking changes

None.

## Verification

- custom-path reproduction changed from failing to generating 7/7 specs
- default configuration path still generates 7/7 specs
- full build, typecheck, lint, and `git diff --check` passed

Fixes #1554

### DIFF
--- a/packages/nitrogen/src/index.ts
+++ b/packages/nitrogen/src/index.ts
@@ -10,6 +10,7 @@ import { promises as fs } from 'fs'
 import { isValidLogLevel, setLogLevel } from './Logger.js'
 import { initNewNitroModule } from './init.js'
 import { NITROGEN_VERSION } from './config/nitrogenVersion.js'
+import { NitroConfig } from './config/NitroConfig.js'
 
 const commandName = 'nitrogen'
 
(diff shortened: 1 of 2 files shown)


packages/nitrogen/src/index.ts
```diff
@@ -57,6 +58,7 @@ await yargs(hideBin(process.argv))
     async (argv) => {
       const basePath = argv.basePath
       const outputDirectory = argv.out
+      NitroConfig.loadFromFile(argv.config)
```
mrousavy, line 61: Nah.
This is not clean at all and hijacks the original loading path internally. Not good, hard to test, easy to go out of sync. It's a race if you forget to call loadFromFile, or use .current first.

mrousavy (changes requested):

PR #1560, opened 2026-08-30
fix: discard impossible Promise listeners

### LINKED ISSUE #1559: Settled C++ Promises retain impossible listeners
## Description

`Promise<T>` and `Promise<void>` enqueue a listener whenever the Promise is not already settled with that listener's matching outcome. Adding a rejection listener to an already-resolved Promise—or a resolution listener to an already-rejected Promise—therefore retains the callback and everything it captures until the Promise itself is destroyed, even though that callback can never run.

This is the C++ Promise implementation; open PR #1441 concerns the separate Android `JPromise` implementation.

## Expected behavior

Listeners should be queued only while a Promise is pending. A listener for an impossible outcome on an already-settled Promise should be discarded immediately.

OskarEichler:
## Summary

- queue C++ Promise listeners only while the Promise is pending
- immediately release opposite-outcome listeners added after settlement
- cover resolved/rejection and rejected/resolution ownership with weak-pointer regressions

## Breaking changes

None. Callbacks that could never execute are no longer retained.

## Verification

- exact Android baseline regression failed; fixed Android and iOS Harness checks pass (1/1 each)
- regenerated specs, full build, post-build typecheck, lint, and C++ format pass
- Android Debug assemble and full iOS Debug Simulator build pass
- `git diff --check` passes

Fixes #1559

### DIFF
--- a/example/src/getTests.ts
+++ b/example/src/getTests.ts
@@ -1554,6 +1554,19 @@ export function getTests(
           .didNotThrow()
           .equals(true)
     ),
+    ...(testObject.name === 'TestObjectCpp'
+      ? [
+          createTest('settled Promises release opposite listeners', () =>
+            it(() =>
+              (
+                testObject as TestObjectCpp
+              ).settledPromiseReleasesOppositeListeners()
+            )
+              .didNotThrow()
+              .equals(true)
+          ),
+        ]
+      : []),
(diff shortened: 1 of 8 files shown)


mrousavy (changes requested):
Why? This adds performance overhead for every call, and chances are very high that isPending is true anyways in that path. So it makes every Promise slower by default for no real benefit?

PR #1562, opened 2026-08-30
fix: guard disposed HybridObject method calls

### LINKED ISSUE #1561: Release HybridFunction calls dereference invalid NativeState
## Description

`HybridFunction` validates the fetched/cast NativeState pointers only under `NITRO_DEBUG`. In release builds, calling a HybridObject method after `dispose()` produces a null shared pointer and then invokes the member on null. Borrowing a method onto a different HybridObject type similarly dereferences a failed dynamic cast. JSI permits a present NativeState slot whose value is null after `setNativeState(nullptr)`.

## Expected behavior

The two post-fetch pointer guards should remain active in every build, producing the existing actionable JSI errors instead of undefined behavior.

OskarEichler:
## Summary

- keep the cheap null and wrong-type NativeState guards active in release builds
- preserve the existing detailed debug-only prechecks
- cover disposed receivers and borrowed methods on the wrong HybridObject type across C++ and Swift/Kotlin implementations

## Breaking changes

None. Invalid method receivers now throw the existing actionable JSI errors consistently in release builds.

## Verification

- root build, typecheck, lint, C++ lint, and `git diff --check` pass
- Android Debug build and targeted Harness pass 4/4
- Android arm64 Release assembles successfully
- iOS Simulator Debug build and targeted Harness pass 4/4

Fixes #1561

### DIFF
--- a/packages/react-native-nitro-modules/cpp/core/HybridFunction.hpp
+++ b/packages/react-native-nitro-modules/cpp/core/HybridFunction.hpp
@@ -220,22 +220,18 @@ private:
 
     // 3. Get `NativeState` from the jsi::Object and check if it is non-null
     std::shared_ptr<jsi::NativeState> nativeState = object.getNativeState(runtime);
-#ifdef NITRO_DEBUG
     if (nativeState == nullptr) [[unlikely]] {
       throw jsi::JSError(runtime, "Cannot " + getHybridFuncDebugInfo<THybrid>(funcKind, funcName) +
                                       " - `this`'s `NativeState` is `null`, "
                                       "did you accidentally call `dispose()` on this object?");
     }
-#endif
 
     // 4. Try casting it to our desired target type.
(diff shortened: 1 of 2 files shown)


packages/react-native-nitro-modules/cpp/core/HybridFunction.hpp
```diff
@@ -220,22 +220,18 @@ class HybridFunction final {
 
     // 3. Get `NativeState` from the jsi::Object and check if it is non-null
     std::shared_ptr<jsi::NativeState> nativeState = object.getNativeState(runtime);
-#ifdef NITRO_DEBUG
```
mrousavy, line 223: Why remove this? Why add the guard in release too? This makes every operation slower by default. For what gain?

mrousavy (changes requested):

PR #1564, opened 2026-08-30
fix: preserve __proto__ record keys

### LINKED ISSUE #1563: Native record conversion loses __proto__ keys
## Description

Native-to-JavaScript conversion for `AnyMap` and typed `Record` values writes keys with ordinary property assignment. A record key named `__proto__` therefore invokes `Object.prototype`'s legacy setter: object values replace the result prototype, while primitive values disappear without creating an own key.

## Expected behavior

`__proto__` should be emitted as a normal own enumerable/writable/configurable record property, preserving its converted value and the result's ordinary Object prototype across C++ and Kotlin/Swift implementations.

OskarEichler:
## Summary

- preserve `__proto__` as record data during native-to-JavaScript AnyMap and typed Record conversion
- keep the output's ordinary Object prototype for object and primitive values
- centralize the exceptional define-property path while retaining cached ordinary property writes
- cover C++ and Kotlin/Swift implementations through both AnyMap and typed map APIs

## Breaking changes

None.

## Verification

- exact Android baseline failed all 4 targeted cases; fixed Android and iOS Harness checks pass 4/4 each
- specs, root build, typecheck, Nitro/example lint, C++ lint, and package Jest pass
- Android Debug build and clean iOS Simulator build pass
- `git diff --check` passes

Fixes #1563

### DIFF
--- a/example/src/getTests.ts
+++ b/example/src/getTests.ts
@@ -960,6 +960,19 @@ export function getTests(
         .didNotThrow()
         .equals(TEST_MAP_4)
     ),
+    createTest('copyAnyMap(...) preserves __proto__ keys', () =>
+      it(() => {
+        const map = { ['__proto__']: { value: 'kept' } }
+        const copy = testObject.copyAnyMap(map)
+        return [
+          Object.prototype.hasOwnProperty.call(copy, '__proto__'),
+          Reflect.get(copy, '__proto__'),
+          Object.getPrototypeOf(copy) === Object.prototype,
+        ]
+      })
+        .didNotThrow()
+        .equals([true, { value: 'kept' }, true])
+    ),
 
     // Test errors
     createTest('funcThatThrows() throws', () =>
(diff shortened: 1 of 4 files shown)


example/src/getTests.ts
```diff
@@ -1454,6 +1467,19 @@ export function getTests(
         .didReturn('object')
         .equals(TEST_MAP_3)
     ),
+    createTest('bounceSimpleMap(map) preserves __proto__ keys', () =>
+      it(() => {
+        const map = { ['__proto__']: 42 }
+        const copy = testObject.bounceSimpleMap(map)
+        return [
+          Object.prototype.hasOwnProperty.call(copy, '__proto__'),
+          Reflect.get(copy, '__proto__'),
```
mrousavy, line 1476: Please give me one single example where you pass `__proto__` to a Nitro func that accepts `AnyMap`. One single example.

mrousavy (changes requested):

PR #1566, opened 2026-08-30
fix: correct Android native registration guidance

### LINKED ISSUE #1565: Generated Android guidance references nonexistent registerNatives()
## Description

Nitrogen generates a `registerAllNatives()` API but its deprecation replacement, JNI_OnLoad example, and `HybridObjectRegistry::createHybridObject()` setup error all instruct users to call namespace-level `registerNatives()`, which does not exist. Following the generated migration or error guidance fails to compile instead of registering HybridObjects.

## Expected behavior

All generated declarations, examples, and runtime setup guidance should consistently reference the generated `registerAllNatives()` function.

OskarEichler:
## Summary

- point the generated initialize deprecation replacement at `registerAllNatives()`
- correct the generated JNI_OnLoad example in both tracked generated headers
- correct HybridObjectRegistry's missing-registration setup error
- add an exact runtime guidance regression across C++ and Kotlin/Swift test suites

## Breaking changes

None.

## Verification

- untouched Android APK failed both guidance assertions; rebuilt APK passes 2/2
- regenerated specs, root build/typecheck, example lint, C++ lint, and `git diff --check` pass
- Android Debug native build passes
- one parallel typecheck/build race was discarded; the clean sequential rerun passed

Fixes #1565

### DIFF
--- a/example/src/getTests.ts
+++ b/example/src/getTests.ts
@@ -2585,6 +2585,24 @@ export function getTests(
         .didNotThrow()
         .equals(true)
     ),
+    createTest(
+      'NitroModules.createHybridObject(...) error names registerAllNatives()',
+      () =>
+        it(() => {
+          try {
+            NitroModules.createHybridObject('__MissingHybridObjectForTest__')
+            return false
+          } catch (error) {
+            return (
+              error instanceof Error &&
+              error.message.includes('registerAllNatives()') &&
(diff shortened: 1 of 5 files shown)


mrousavy (approved):
Interesting find - I thought I had that replaced. Let's see CI run

PR #1568, opened 2026-08-30
fix: validate JavaScript enum values

### LINKED ISSUE #1567: Generated numeric enum conversion accepts invalid JavaScript values
## Description

Nitrogen generates a `canConvert()` validator for numeric enums, but direct generated `fromJSI()` conversion never calls it. Direct enum properties therefore accept undeclared, fractional, and non-finite JavaScript numbers and cast them to the closed C++ enum. The validator also casts non-finite and out-of-range doubles to `int` before it can reject them.

## Expected behavior

Generated direct enum conversion should reject every value that is not one of the declared integer members, and validation should reject unsafe numbers before converting them to `int`.

OskarEichler:
## Summary

- validate direct generated numeric-enum conversions before casting
- reject NaN, infinities, and out-of-range values before converting a JavaScript number to `int`
- cover invalid direct enum assignments through both C++ and Swift/Kotlin TestObject implementations
- commit the regenerated enum headers

## Breaking changes

Invalid JavaScript numbers that were previously cast into undeclared closed-enum values now throw during conversion. Declared numeric enum values are unchanged.

## Verification

- exact untouched iOS Harness rejected 0/5 invalid values in both native implementation paths; fixed regression passes 2/2
- focused enum suite passes 22/22
- full root build, workspace typecheck, lint-all, package Jest, native iOS build, Android arm64 Debug build, and `git diff --check` pass

Fixes #1567

### DIFF
--- a/example/src/getTests.ts
+++ b/example/src/getTests.ts
@@ -521,6 +521,22 @@ export function getTests(
         .didReturn(typeof OldEnum.SECOND)
         .equals(OldEnum.SECOND)
     ),
+    createTest('set optionalOldEnum rejects invalid numbers', () =>
+      it(() => {
+        const invalidValues = [9999, 1.5, NaN, Infinity, -Infinity]
+        let rejectedValues = 0
+        for (const value of invalidValues) {
+          try {
+            testObject.optionalOldEnum = value as OldEnum
+          } catch {
+            rejectedValues++
+          }
+        }
+        return rejectedValues
+      })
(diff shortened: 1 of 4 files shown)


mrousavy (changes requested):
This makes every enum parse slower. 
Why? What is the gain? Your example uses an untyped force cast in TypeScript, basically making it non-typed. I don't think this is something we should support

PR #1570, opened 2026-08-30
fix: support aliased HybridView imports

### LINKED ISSUE #1569: Nitrogen rejects aliased HybridView imports
## Description

Nitrogen semantically identifies a type using `HybridView as SomeAlias` as a Hybrid View, but platform extraction then checks the source spelling against the literal name `HybridView`. Generation fails with an error claiming the detected type does not alias `HybridView<...>`.

## Expected behavior

Hybrid View platform extraction should use the resolved `HybridViewTag<PlatformSpec>` type, so valid import aliases retain the same default or explicit platform configuration as a directly named `HybridView`.

OskarEichler:
## Summary

- derive Hybrid View platforms from the resolved `HybridViewTag<PlatformSpec>` intersection
- support import aliases without depending on the source-level spelling `HybridView`
- preserve default and explicit platform selections
- cover generation through an aliased TestView import

## Breaking changes

None.

## Verification

- exact baseline fixture exited 1 because the aliased type was detected as a Hybrid View and then rejected during platform extraction
- fixed direct regeneration passes 7/7 with zero tracked generated drift
- explicit `{ ios: "c++" }` semantic probe resolves correctly
- full root build, sequential workspace typecheck, lint-all, and `git diff --check` pass
- one parallel typecheck/build race was discarded; the clean sequential run passed

Fixes #1569

### DIFF
--- a/packages/nitrogen/src/getPlatformSpecs.ts
+++ b/packages/nitrogen/src/getPlatformSpecs.ts
@@ -1,6 +1,6 @@
 import type { PlatformSpec } from 'react-native-nitro-modules'
 import type { InterfaceDeclaration, Type, TypeAliasDeclaration } from 'ts-morph'
-import { Node, Symbol } from 'ts-morph'
+import { Symbol } from 'ts-morph'
 import { getBaseTypes } from './utils.js'
 
 export type Platform = keyof Required<PlatformSpec>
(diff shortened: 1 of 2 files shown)


packages/nitrogen/src/getPlatformSpecs.ts
```diff
@@ -185,27 +185,22 @@ export function getHybridObjectPlatforms(
 export function getHybridViewPlatforms(
```
mrousavy, line 185: Does this only affect Hybrid Views? What about Hybrid Objects?

OskarEichler: Only Hybrid Views were affected. `getHybridObjectPlatforms()` already resolves the semantic base type via `getBaseTypes(...)` and `isDirectlyType(...)`, which checks both the symbol and alias symbol. The Hybrid View path was the remaining syntax-based check (`getTypeName().getText() === "HybridView"`), so an aliased import only failed there.

mrousavy: Cool thanks!

PR #1590, opened 2026-09-05
feat: add discriminated union types to nitrogen

### LINKED ISSUE #1466: Add Discriminating Types
### Feature Request/Enhancement

We should add Discriminating Types to Nitro(gen).

The design requires;
- A union type
- Each type in the union types must have one literal that discriminates it from the rest

For example;

```ts
interface Bike {
  type: 'bike'
  canDoWheelie: boolean
}
interface Car {
  type: 'car'
  canDrift: boolean
}
type Vehicle = Bike | Car
```

Here, `type` is the discriminator, if it is the string literal `'bike'` at runtime, it has to be a `Bike` struct. If it is `'car'`, it has to be a `Car` struct.

Note: with different non-nullable properties like `canDoWheelie` and `canDrift` in the example above we don't even need the discriminator `type`, we can also just check if `canDoWheelie` is contained and if yes it's a `Bike`, **however** this will not work if the properties are nullable, or overlapping.

For example, this surfaced during react-native-vision-camera development:

```ts
type AudioCategory = 'play' | 'record' | 'play-and-record'

interface AutomaticAudioSessionConfiguration {
  allowBackgroundAudio?: boolean
}
interface ManualAudioSessionConfiguration {
  allowBackgroundAudio?: boolean
  category?: AudioCategory
}
type AudioConfiguration = AutomaticAudioSessionConfiguration | ManualAudioSessionConfiguration
```

Here, at runtime we do not know if the user wants to select `AutomaticAudioSessionConfiguration` or `ManualAudioSessionConfiguration` **when all properties are null**. There is no difference.
Hence, the user has to specify intent via a discriminating union:

```diff
  type AudioCategory = 'play' | 'record' | 'play-and-record'

  interface AutomaticAudioSessionConfiguration {
+   type: 'automatic'
    allowBackgroundAudio?: boolean
  }
  interface ManualAudioSessionConfiguration {
+   type: 'manual'
    allowBackgroundAudio?: boolean
    category?: AudioCategory
  }
  type AudioConfiguration = AutomaticAudioSessionConfiguration | ManualAudioSessionConfiguration
```

Implementation is quite tricky, as literals have to carry context based on if they are in a union or not, and then we need to generate an extra enum and extra C++ `JSIConverter` overload.

Also, I guess this doesn't apply to `HybridObject`s, only flat structs.

### Are there any existing workarounds?

Non-nullable properties on the structs

### Additional information

- [x] I checked the [documentation](https://nitro.margelo.com/docs/) and couldn't find this feature.
- [ ] I can create a PR to implement this feature. (See [Contributing](https://nitro.margelo.com/docs/contributing) for more information)

dhruvpatel374:
Closes #1466

## What

Adds support for discriminated union types in nitrogen. When a TypeScript union consists of interfaces that share a common string literal property, nitrogen now detects the discriminant key and generates a named `JSIConverter` that dispatches on it at runtime instead of falling back to positional `std::variant` indexing.

```ts
interface Truck { kind: 'truck'; payload: number }
interface Boat  { kind: 'boat';  lengthMeters: number }
type Vehicle = Truck | Boat
```

Here `kind` is the discriminant. At runtime, if `kind` is `"truck"` the converter deserializes a `Truck`, if `"boat"` it deserializes a `Boat`. This also solves the ambiguous-all-nulls problem from the issue - even when every other property is optional/null, the discriminant tells the converter exactly which struct to construct.

## How it works

- `createType.ts` - added `findDiscriminantKey()` that inspects a union's constituents for a shared property where every member has a unique string literal value. If found, returns a `DiscriminatedUnionType` instead of a plain `VariantType`.
- `getInterfaceProperties.ts` - added optional `skipKeys` param so the discriminant property is excluded from the struct fields (the union's `JSIConverter` handles dispatch, the struct itself doesn't need it).
- `CppDiscriminatedUnion.ts` - generates a `JSIConverter<std::variant<A, B>>` specialization that reads the discriminant key first, then delegates to each struct's own converter.
- `SwiftDiscriminatedUnion.ts` - generates a `@frozen enum` with cases named after discriminant values (e.g. `.truck`, `.boat`) instead of positional `.first`, `.second`.
- `KotlinDiscriminatedUnion.ts` - generates a `sealed class` with inner classes named after discriminant values.
- Wired up in `SwiftCxxBridgedType`, `KotlinCxxBridgedType`, `SwiftCxxTypeHelper`, and `getReferencedTypes`.

## Generated output

**C++** (`Vehicle.hpp`):
```cpp
template <>
struct JSIConverter<std::variant<Truck, Boat>> final {
  static inline std::variant<Truck, Boat> fromJSI(...) {
    // reads "kind", switches on hashString, delegates to Truck or Boat converter
  }
};
```

**Swift** (`Vehicle.swift`):
```swift
@frozen public indirect enum Vehicle {
  case truck(Truck)
  case boat(Boat)
}
```

**Kotlin** (`Vehicle.kt`):
```kotlin
sealed class Vehicle {
  data class Truck(val value: Truck): Vehicle()
  data class Boat(val value: Boat): Vehicle()
}
```

## Test

- Added `Truck`, `Boat`, `Vehicle` (discriminant key `kind`) to `TestObject.nitro.ts`
- Added `bounceVehicle` method to C++, Swift, and Kotlin test implementations
- Added runtime assertions in `getTests.ts` covering both the `truck` and `boat` cases

## Checklist

- [x] `bun specs` run and generated files committed
- [x] `bun lint-all` passes
- [x] `bun typecheck` passes
- [x] No existing specs or test cases removed
- [x] Diff is scoped to this feature only


dhruvpatel374:
Hey @mrousavy , just wanted to check in happy to make any changes if needed! 🙏

PR #1591, opened 2026-09-05
fix(nitrogen): look up Kotlin callbacks with a boxed return type

giaBaoJS:
## What breaks

[`packages/nitrogen/src/syntax/kotlin/KotlinFunction.ts:142`](https://github.com/margelo/nitro/blob/main/packages/nitrogen/src/syntax/kotlin/KotlinFunction.ts#L142) builds the JNI signature of the generated `JFunc_X::invoke()` from the plain JNI type of the callback's return type. For `Sync<() => number>` that is `double()`:

```cpp
double invoke() const {
  static const auto method = javaClassStatic()->getMethod<double()>("invoke");
  auto __result = method(self());
  return __result;
}
```

The Kotlin half of the same callback is a `fun interface` that extends the Kotlin function type:

```kotlin
fun interface Func_double: () -> Double {
  override fun invoke(): Double
}
```

`invoke` overrides `Function0<R>.invoke`, and kotlinc cannot specialize a generic return type to a JVM primitive, so it keeps the boxed type. Parameters are specialized (kotlinc adds a `(Ljava/lang/Object;)Ljava/lang/Object;` bridge for them), return types are not.

fbjni resolves methods by exact descriptor, so the Android build does not catch this. `GetMethodID` finds nothing and the call throws `NoSuchMethodError`.

## Ground truth

kotlinc 2.4.10 + `javap -p -s`, on the exact `fun interface` shape nitrogen generates:

| Kotlin declaration | what kotlinc emits | what nitrogen looked up |
| --- | --- | --- |
| `fun interface F: () -> Double` | `invoke` `()Ljava/lang/Double;` | `()D` |
| `fun interface F: () -> Boolean` | `invoke` `()Ljava/lang/Boolean;` | `()Z` |
| `fun interface F: () -> Long` | `invoke` `()Ljava/lang/Long;` | `()J` |
| `fun interface F: () -> Unit` | `invoke` `()V` | `()V` (already correct) |
| `fun interface F: (Double) -> Unit` | `invoke` `(D)V` | `(D)V` (already correct) |
| `fun interface F: () -> String` | `invoke` `()Ljava/lang/String;` | same (already correct) |

`Unit` is the exception: it maps to `void`, so a void callback was fine. Every other primitive return was broken.

## When it fires

Only when the callback was implemented in Kotlin. A callback that came from JS is a `Func_X_cxx`, and `KotlinCxxBridgedType` unwraps that via `getFunction()` without ever touching JNI. Anything else falls into `JNICallable<JFunc_X, R(Args...)>`, which calls `JFunc_X::invoke()`.

That is why CI never hit it: `react-native-nitro-test` used `Func_double` only as a parameter (`callbackSync`), never as a value handed from Kotlin to C++.

iOS is unaffected, Swift closures do not go through this path.

## Fix

`KotlinFunction.ts` now boxes the return type of the `invoke` lookup for the primitive kinds, and unboxes `__result` with the existing `parseFromKotlinToCpp(.., isBoxed)` machinery:

```cpp
double invoke() const {
  static const auto method = javaClassStatic()->getMethod<jni::local_ref<jni::JDouble>()>("invoke");
  auto __result = method(self());
  return __result->value();
}
```

`std::optional<double>` returns already emitted `jni::local_ref<jni::JDouble>`, so they were correct before and are untouched.

## Test

One method on `SharedTestObjectProps`, in the "Sync funcs" block next to the existing `callbackSync`:

```ts
getSyncNumberCallback(): Sync<() => number>
```

plus one assertion in `example/src/getTests.ts`, so the Harness workflows cover it on both platforms. It reuses the `Func_double` specialization that `callbackSync` already generates, this time in the return direction, which is the direction that was broken.

## How I proved it

I compiled the generated Kotlin (`nitrogen/generated/android/kotlin`, minus the ViewManagers, which need React Native on the classpath) with kotlinc, then checked every `getField` / `getMethod` lookup in `nitrogen/generated/android/c++` against `javap -p -s` on the resulting class files.

Before the fix, one lookup did not exist:

```
('JFunc_double.hpp', 'com.margelo.nitro.test.Func_double', 'invoke', 'Method', '()D', ['()Ljava/lang/Double;', '()Ljava/lang/Object;'])
1 signature problems
```

Resolving that descriptor the way `GetMethodID` does:

```
MISSING  com.margelo.nitro.test.Func_double.invoke()D
FOUND    com.margelo.nitro.test.Func_double.invoke()Ljava/lang/Double;
```

After the fix, all 339 JNI names and descriptors emitted for `react-native-nitro-test` match the compiler, with no exceptions:

```
0 signature problems
```

I also checked the fix is not vacuous by mutating it in the other direction (boxing to `jni::JLong` instead of `jni::JDouble`), which the same check catches as `()Ljava/lang/Long;` vs `()Ljava/lang/Double;`.

## Not covered

`UInt64` callbacks are broken in a different way and are out of scope here: `ULong` is a value class, so kotlinc mangles the whole method name (`invoke-s-VKNKU` for a `ULong` return, `invoke-VKZWuLQ` for a `ULong` parameter). Boxing does not help there, so I left that path exactly as it was. Happy to open a separate issue for it.

## Checks

* `bun specs` run, generated files committed. Apart from the new method, the only generated change is `JFunc_double.hpp` and some include reordering.
* `bun run build`, `bun typecheck`, `bun lint`, `bun lint-cpp`, `bun lint-swift`, `bun lint-kotlin`: all clean, no files changed by the formatters.
* I could not run the Harness suite here (no Android device), so the new runtime assertion is verified by CI, not locally. The compiler evidence above is what I verified locally.


packages/nitrogen/src/syntax/kotlin/KotlinFunction.ts
```diff
@@ -139,7 +158,10 @@ class ${name}_java(private val function: ${lambdaSignature}): ${name} {
       return bridge.parseFromCppToKotlin(p.escapedName, 'c++', false)
     }),
   ]
-  const jniSignature = `${bridgedReturn.asJniReferenceType('local')}(${functionType.parameters
+  const jniReturnType = isReturnBoxed
+    ? `jni::local_ref<${bridgedReturn.getTypeCode('c++', true)}>`
+    : bridgedReturn.asJniReferenceType('local')
+  const jniSignature = `${jniReturnType}(${functionType.parameters
```
mrousavy: What? This does not make sense and looks like AI slop. `.asJniReference('local')` gives you `jni::local_ref<...>`, so this ternary here adds exactly what the method already does, just more complex code. Was this generated by GPT-1 on ultra-low?

mrousavy (changes requested):
code is bad, pls fix

mrousavy (changes requested):
~

mrousavy (changes requested):
~

packages/nitrogen/src/syntax/kotlin/KotlinFunction.ts
giaBaoJS: Simplified. The boxing decision moved into `KotlinCxxBridgedType`, so the call site is now a single `asJniReferenceType('local', isReturnBoxed)` and the hand-built `jni::local_ref<...>` string is gone. Generated output is byte-identical.

On the redundancy: `asJniReferenceType` returns a bare primitive for exactly the kinds this PR is about.

```ts
case 'void':
case 'number':
case 'boolean':
case 'int64':
  // primitives are not references
  return this.getTypeCode('c++')
```

So `() => number` produced `getMethod<double()>("invoke")`, i.e. `()D`. The compiled `fun interface Func_double: () -> Double` only has:

```
invoke  ->  ()Ljava/lang/Double;
invoke  ->  ()Ljava/lang/Object;
```

Parameters do specialize, `(Double) -> Double` compiles to `(D)Ljava/lang/Double;`. Only the return stays generic.

Separate finding while measuring: `UInt64` returns are broken too, but boxing does not fix them. `() -> ULong` compiles to `invoke-s-VKNKU -> ()J` plus the `()Ljava/lang/Object;` bridge, so there is no method named `invoke` to resolve at all. Kept out of this PR. Happy to add a nitrogen error for it if you want.

packages/nitrogen/src/syntax/kotlin/KotlinFunction.ts
```diff
@@ -110,6 +110,9 @@ class ${name}_java(private val function: ${lambdaSignature}): ${name} {
     `${name}_cxx`
   )
   const bridgedReturn = new KotlinCxxBridgedType(functionType.returnType)
+  // `invoke` overrides `FunctionN.invoke`, so its return type is a generic - unlike
+  // its parameters, which kotlinc specializes to primitives.
+  const isReturnBoxed = bridgedReturn.isBoxedAsJvmGeneric
```
mrousavy: I wonder if we even need this information. Can't we just always pass `true` to `asJniReferenceType(..., true)`?

I guess if it's `void` it shouldn't generate `jni::local_ref<void>`, but that logic could be handled inside `asJniReferenceType(...)` too right? Like returning `void` for `'void'` in the switch...

mrousavy (changes requested):

PR #1640, opened 2026-09-07
fix: honor Nitrogen config path

### LINKED ISSUE #1554: Nitrogen ignores the --config option

huytdps13400:
## Summary

- pass the parsed --config value into Nitrogen before generation starts
- centralize file-backed singleton initialization in NitroConfig.load() while preserving ./nitro.json as the default
- add a black-box CLI regression and run Nitrogen CLI tests in CI

## Breaking changes

None. Existing invocations still load ./nitro.json by default.

## Verification

- regression test failed before the fix: the CLI logged Loading nitro.json and reported that ./nitro.json did not exist despite a valid --config path
- bun --cwd packages/nitrogen test (1/1)
- bun --cwd packages/nitrogen typecheck
- bun --cwd packages/nitrogen lint-ci
- bun run build
- bun typecheck
- bun specs, with no generated diff
- C++ and Swift format scripts passed through the Xcode toolchain
- full lint-all reached Kotlin lint but the local machine does not have the ktlint executable; no Kotlin/native files are changed and CI will run that lane on the configured runner

## AI assistance

Codex using GPT-5.6 Sol assisted with reproduction, implementation, tests, and review.

Fixes #1554

### DIFF
--- a/packages/nitrogen/src/index.ts
+++ b/packages/nitrogen/src/index.ts
@@ -10,6 +10,7 @@ import { promises as fs } from 'fs'
 import { isValidLogLevel, setLogLevel } from './Logger.js'
 import { initNewNitroModule } from './init.js'
 import { NITROGEN_VERSION } from './config/nitrogenVersion.js'
+import { NitroConfig } from './config/NitroConfig.js'
 
 const commandName = 'nitrogen'
 
@@ -57,6 +58,7 @@ await yargs(hideBin(process.argv))
     async (argv) => {
       const basePath = argv.basePath
       const outputDirectory = argv.out
+      NitroConfig.load(argv.config)
       await runNitrogenCommand(basePath, outputDirectory)
     }
   )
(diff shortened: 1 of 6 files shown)


packages/nitrogen/src/index.ts
```diff
@@ -57,6 +58,7 @@ await yargs(hideBin(process.argv))
     async (argv) => {
       const basePath = argv.basePath
       const outputDirectory = argv.out
+      NitroConfig.load(argv.config)
```
mrousavy, line 61: that's not safe and just prone to races.

mrousavy (changes requested):

PR #1655, opened 2026-09-21
fix(views): fill Props::rawProps on Android when overriding cloneProps

### LINKED ISSUE #1656: Android: Hybrid Views receive no base ViewProps on React Native >= 0.86
### What happened?

`ViewComponentDescriptor::cloneProps` overrides React Native's implementation and
does not call `initializeDynamicProps`. Up to RN 0.85 that was harmless, because
`Props::Props` called it for you via `Props::initialize`. RN 0.86 removed that, so
nothing fills `Props::rawProps` for a Hybrid View any more — and on Android that
map is exactly what gets serialized to Java and applied by
`ViewManager.updateProperties`.

The effect: a Hybrid View silently receives **no** `backgroundColor`, `border*`,
`transform`, `opacity`, `testID` or accessibility props on Android. iOS is
unaffected, because `RCTViewComponentView` reads the typed C++ fields directly.

Nothing warns, because React Native never sees a prop to report as unsupported.

Left is a plain RN `View`, right is `TestView`, **same style object**, on RN 0.87.1
with no feature flags and an otherwise unmodified example app:

before and after

`opacity` is the prop to watch: `TestView` paints its own surface, which masks
whether `backgroundColor` arrived, but nothing native can override a view's alpha.

**Why it happens.** `Props::rawProps` has exactly one writer per RN version:

| | writers of `Props::rawProps` |
| --- | --- |
| RN 0.85.3 | `Props::initialize` (from the **constructor**) + `ConcreteComponentDescriptor::cloneProps` |
| RN 0.87.1 | `ConcreteComponentDescriptor::cloneProps` **only** |

(`ShadowNode::propsForClonedShadowNode` also touches it, but it is gated on
`!rawProps.empty()` — it merges, it cannot create.)

Android then does, in `SurfaceMountingManager.kt`:

```kotlin
viewState.currentProps = ReactStylesDiffMap(props)
viewManager.updateProperties(view, viewState.currentProps)
```

An empty `rawProps` means an empty `ReactStylesDiffMap`, so every `@ReactProp`
setter on `BaseViewManager` is skipped.

One corroborating detail: nitrogen generates a per-view `filterObjectKeys` and
threads it through `ViewProps → BaseViewProps → YogaStylableProps → Props`, where
React Native's only consumer of it is `rawProps.toDynamic(filterObjectKeys)` inside
`initializeDynamicProps`. Without that call it is dead code.

**Suggested fix** (this is what #1655 does):

```cpp
auto newProps = TShadowNode::Props(context, /* & */ rawProps, props);
#ifdef RN_SERIALIZABLE_STATE
  TShadowNode::initializeDynamicProps(newProps, rawProps, props);
#endif
return newProps;
```

It does add a `rawProps.toDynamic()` per props clone, which may be part of what the
override was avoiding — if you'd prefer something narrower (serializing only base
`ViewProps` keys), happy to rework it.

**Why the harness never caught this.** It couldn't: on the pinned 0.85.3 the
constructor still fills the map, so the bug has no observable effect. With
`enableExclusivePropsUpdateAndroid` on — the path 0.86+ takes unconditionally — it
does, and not subtly. `queryByTestId` resolves against the native view tree and
`testID` is itself a base view prop, so most of the file can't find the view it
just rendered:

```
full nitro.views.harness, Pixel 9 API 36 emulator, flag on

  without the fix ....  8 failed,  2 passed
  with it ............  10 passed
```

Six of those eight are existing tests.

**Reproductions**, both off `b06f548` and buildable as-is:

| branch | what it is |
| --- | --- |
| [`bump/rn-0.87`](https://github.com/ronickg/nitro/tree/bump/rn-0.87) | RN 0.87.1 + the two boxes above on the example's **View** tab. No fix, no flags. Build it, open the View tab — that is the screenshot. |
| #1655 | the fix, the flag that lets the harness exercise it, and tests |

### Reproduceable Code

const box = {
  width: 120,
  height: 120,
  backgroundColor: '#00C000',
  transform: [{ rotate: '10deg' }],
  opacity: 0.4,
} as const

// Same style object on both. On Android the Nitro view ignores all of it.
<View style={box} />
<TestView style={box} isBlue={false} hasBeenCalled={false} colorScheme="light" />

### Relevant log output

# ViewComponentDescriptor::cloneProps instrumented to print newProps->rawProps.size()

example on RN 0.85.3, flag off (today's default) ........  5
example on RN 0.85.3, enableExclusivePropsUpdateAndroid .  0   (13,405 consecutive clones)
example on RN 0.87.1, unpatched .........................  0   (6,508 consecutive clones)
our own app on RN 0.87.1, patched .......................  102

# Nothing is logged by React Native in the broken case - an empty rawProps means
# an empty ReactStylesDiffMap, so there is no unsupported prop for it to warn about.

### Device

Pixel 9 API 36 emulator (Android 16, arm64). Not tested on physical hardware; the iOS claim above is from reading `RCTViewComponentView`, not a measurement.

### Nitro Modules Version

0.37.1, and `main` (b06f548)

### Nitrogen Version

0.37.1

### Additional information

- [ ] I am using Expo
- [X] I am using nitrogen
- [X] I have read and followed the [Troubleshooting Guide](https://nitro.margelo.com/docs/troubleshooting).
- [X] I created a reproduction PR to reproduce this issue here in the nitro repo. (See [Contributing](https://nitro.margelo.com/docs/contributing) for more information)
- [X] I searched for [similar issues in this repository](https://github.com/margelo/nitro/issues) and found none.

ronickg:
Fixes #1656.

## The problem

`ViewComponentDescriptor::cloneProps` overrides React Native's implementation and
does not call `initializeDynamicProps`. Up to RN 0.85 that was harmless, because
`Props::Props` called it for you via `Props::initialize`. RN 0.86 removed that, so
nothing fills `Props::rawProps` for a Hybrid View any more — and on Android that
map is exactly what gets serialized to Java and applied by
`ViewManager.updateProperties`.

The effect: a Hybrid View silently receives **no** `backgroundColor`, `border*`,
`transform`, `opacity`, `testID` or accessibility props on Android. iOS is
unaffected, because `RCTViewComponentView` reads the typed C++ fields directly.

Nothing warns, because React Native never sees a prop to report as unsupported.

Left is a plain RN `View`, right is `TestView`, **same style object**, on RN 0.87.1
with no feature flags:

before and after

`Props::rawProps` has exactly one writer per RN version:

| | writers of `Props::rawProps` |
| --- | --- |
| RN 0.85.3 | `Props::initialize` (from the **constructor**) + `ConcreteComponentDescriptor::cloneProps` |
| RN 0.87.1 | `ConcreteComponentDescriptor::cloneProps` **only** |

(`ShadowNode::propsForClonedShadowNode` also touches it, but it is gated on
`!rawProps.empty()` — it merges, it cannot create.)

A corroborating detail: nitrogen generates a per-view `filterObjectKeys` and threads
it through `ViewProps → BaseViewProps → YogaStylableProps → Props`, where React
Native's only consumer of it is `rawProps.toDynamic(filterObjectKeys)` inside
`initializeDynamicProps`. Without that call it is dead code.

## The fix

```cpp
auto newProps = TShadowNode::Props(context, /* & */ rawProps, props);
#ifdef RN_SERIALIZABLE_STATE
  TShadowNode::initializeDynamicProps(newProps, rawProps, props);
#endif
return newProps;
```

This does add a `rawProps.toDynamic()` per props clone, which may be part of what
the override was avoiding. If you'd prefer something narrower — serializing only
the base `ViewProps` keys — I'm happy to rework it that way.

## Why the harness didn't catch it

It couldn't. On the pinned RN 0.85.3 the constructor still fills the map, so the
bug has no observable effect. Turning on `enableExclusivePropsUpdateAndroid` makes
0.85 take the same path 0.86+ takes unconditionally, and then it does — and not
subtly. `queryByTestId` resolves against the native view tree, and `testID` is
itself a base view prop, so most of the file can't even find the view it rendered:

```
full nitro.views.harness, Pixel 9 API 36 emulator, flag on

  without this fix ....  8 failed,  2 passed
  with it .............  10 passed
```

Six of those eight are existing tests, not ones added here:

```
● TestView › renders with native props, layout, pixels, methods, and callbacks
● TestView › updates every prop, changes pixels, and resizes the same native view
● TestView › unmounts and creates a fresh native view when remounted
● multiple RecyclableTestViews › keeps instances isolated while one is updated and recycled
● RecyclableTestView › renders, changes pixels, and resizes the same native view
● RecyclableTestView › reuses and resets the native view across recycling cycles
```

## What's in this PR

| file | why |
| --- | --- |
| `cpp/views/ViewComponentDescriptor.hpp` | the fix |
| `apps/example/android/.../MainApplication.kt` | turns on `enableExclusivePropsUpdateAndroid` so the harness exercises the 0.86+ path. Droppable once the floor is >= 0.86, where it's unconditional. |
| `apps/example/__tests__/nitro.views.harness.tsx` | a dedicated pair of tests naming the cause (`applies opacity to the native view`, plus a control) |

The opacity test asserts the blue surface washes out at 10% alpha rather than
asserting on PNG alpha — a screen capture is always alpha 255, so that would
have passed either way.

## Verification

- `bun specs` — no codegen drift
- `bun typecheck` — clean
- `bun example lint-ci` — clean
- `clang-format -style=file:./config/.clang-format` on the changed header — clean
- `bun lint-kotlin` — ktlint not installed locally; its scope is `packages/*/android`, and the Kotlin change here is under `apps/example`
- `nitro.harness` (the other file) shows 11 pre-existing ArrayBuffer/HardwareBuffer
  failures on this emulator — identical with and without the flag, verified with a
  control build

A clean, fix-free reproduction on RN 0.87.1 is on
[`bump/rn-0.87`](https://github.com/ronickg/nitro/tree/bump/rn-0.87): build it and
open the example's **View** tab — that's the screenshot above.

Tested on a Pixel 9 API 36 emulator (Android 16, arm64). I have not run this on
physical hardware or on iOS; the iOS claim above is from reading
`RCTViewComponentView`, not from a measurement.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


hannojg:
Hey, thanks for finding & working on this!

I think while this is directionally right, there are two problems with that:

- placing the call in the ViewDescriptor's `cloneProps` isn't 100% correct, more on that below
- we should version gate this, so on versions < 0.87 we don't call`initializeDynamicProps` twice unnecessarily

In `ViewComponentDescriptor::cloneProps` we don't have access to the prop's `filterObjectKeys`, thus you call `initializeDynamicProps` without it. This means that:
- react native will try to convert our nitro props to folly::dynamic values, which is unnecessary overhead 
- i believe we could even crash when react native tries to do that, for example for props such as `BitInt` which crash in react-native's folly conversion

So, i think we have to move this to the nitrogen generated props:
- Make sure to version gate it properly
- Make sure to pass `filterObjectKeys`
- (A good place for this could be in RawPropsCompat.cpp as we already have version gated prop logic there)

Also, can we remove the feature flag for the example app? i get that we need to enable it to reproduce the issue below RN 0.87. But maybe we can just set it to false and add a comment that this should be enabled to reproduce this issue etc.

Are you available to update / work on this? Otherwise i can also make the change, just lmk 😊  thanks again for raising this!

hannojg (changes requested):
(See comment above)

ronickg:
Hey @hannojg  Thanks for the quick reply. I think it would be best if you can tackle it. I just wanted to point out the issue which I found and let claude propose a potenial fix.

